PR #4, opened 2021-06-28
Add function to estimate dynamic memory usage of frontiers.

nuttycom:


src/bridgetree.rs
```diff
@@ -301,6 +302,19 @@ impl<H, const DEPTH: u8> Frontier<H, DEPTH> {
     pub fn position(&self) -> Option<Position> {
         self.frontier.as_ref().map(|f| f.position)
     }
+
+    pub fn dynamic_memory_usage(&self) -> usize {
+        size_of::<Option<NonEmptyFrontier<H>>>()
+            + self.frontier.as_ref().map_or(0, |f| {
+                2 * size_of::<usize>() + f.parents.capacity() * size_of::<H>()
+            })
```
str4d, line 310: The first part is not dynamic memory usage, e.g. `self` could be on the stack. Only the caller knows whether this should be included in its own `dynamic_memory_usage` call, and it would figure that out with a call to `size_of::<Frontier<H, DEPTH>>()`.
```suggestion
        self.frontier.as_ref().map_or(0, |f| {
            2 * size_of::<usize>() + f.parents.capacity() * size_of::<H>()
        })
```
Technically this would also undercount if `H` itself allocated memory. I think it's fine to just document that limitation, since this struct only really makes sense with `H` types that are "simple".

src/bridgetree.rs
```diff
@@ -301,6 +302,19 @@ impl<H, const DEPTH: u8> Frontier<H, DEPTH> {
     pub fn position(&self) -> Option<Position> {
         self.frontier.as_ref().map(|f| f.position)
     }
+
+    pub fn dynamic_memory_usage(&self) -> usize {
```
str4d, line 306: Document.

str4d (changes requested):

src/bridgetree.rs
```diff
@@ -294,13 +295,30 @@ pub struct Frontier<H, const DEPTH: u8> {
...
+    /// if the frontier is nonempty.
     pub fn position(&self) -> Option<Position> {
         self.frontier.as_ref().map(|f| f.position)
     }
+
+    /// Return the amount of memory dynamically allocated for parent
+    /// values within the frontier.
+    pub fn dynamic_memory_usage(&self) -> usize {
+        self.frontier.as_ref().map_or(0, |f| {
+            2 * size_of::<usize>() + f.parents.capacity() * size_of::<H>()
```
str4d, line 313: Rust doesn't guarantee that it lays out structs transparently.
```suggestion
            2 * size_of::<usize>() + f.parents.capacity() * size_of::<Parent<H>>()
```

str4d (approved):

PR #5, opened 2021-07-02
Rename "parent" to "ommer" and add public operations required for serialization support.

nuttycom:
The term "parent" as it has previously been used in this module is
confusing, because it referred not to the actual parent of a leaf note,
but to the *sibling* of the leaf's future parent. "Ommer" is a
gender-neutral term for the sibling of a parent, so it is ideal
for use in this context.

### DIFF
--- a/src/bridgetree.rs
+++ b/src/bridgetree.rs
@@ -1,7 +1,7 @@
-/// A space-efficient implementation of the `Tree` interface.
-///
-/// In this module, the term "ommer" is used as a gender-neutral term for
-/// the sibling of a parent node in a binary tree.
+//! A space-efficient implementation of the `Tree` interface.
+//!
+//! In this module, the term "ommer" is used as a gender-neutral term for
+//! the sibling of a parent node in a binary tree.
 use serde::{Deserialize, Serialize};
 
 use std::collections::HashMap;
@@ -28,7 +28,8 @@ impl Position {
     }
 
     /// Returns the altitude of the top of a binary tree containing
-    /// `self + 1` nodes.
+    /// a number of nodes equal to the next power of two greater than 
+    /// or equal to `self + 1`.
     fn max_altitude(&self) -> Altitude {
         Altitude(if self.0 == 0 {
             0
@@ -50,10 +51,15 @@ impl Position {
             })
     }
 
+    /// Returns the number of ommers required to construct an authentication
+    /// path to the root of a merkle tree that has `self + 1` nodes.
     pub fn altitudes_required_count(&self) -> usize {
         self.altitudes_required().count()
     }
 
+    /// Returns the altitude of each cousin and/or ommer required to construct 
+    /// an authentication path to the root of a merkle tree of depth `self + 1` 
+    /// nodes.
     pub fn altitudes_required(&self) -> impl Iterator<Item = Altitude> + '_ {
         (0..=(self.max_altitude() + 1).0)
             .into_iter()
@@ -66,6 +72,9 @@ impl Position {
             })
     }
 
+    /// Returns the altitude of each cousin and/or ommer required to construct 
+    /// an authentication path to the root of a merkle tree containing 2^64
+    /// nodes. 
     pub fn all_altitudes_required(&self) -> impl Iterator<Item = Altitude> + '_ {
         (0..64).into_iter().filter_map(move |i| {
             if self.0 == 0 || self.0 & (1 << i) == 0 {
@@ -76,6 +85,10 @@ impl Position {
         })
     }
 
+    /// Returns whether the binary tree having `self` as the position of the
+    /// rightmost leaf contains a perfect balanced tree of height 
+    /// `to_altitude + 1` that contains the aforesaid leaf, without requiring
+    /// any empty leaves or internal nodes.
     pub fn is_complete(&self, to_altitude: Altitude) -> bool {
         for i in 0..(to_altitude.0) {
             if self.0 & (1 << i) == 0 {
@@ -83,11 +96,6 @@ impl Position {
             }
         }
         true
-    }
-
-    pub fn has_observed(&self, altitude: Altitude, since: Position) -> bool {
-        let altitude_delta = 2usize.pow(altitude.0.into());
-        self.0 - since.0 > altitude_delta
     }
 }
 
@@ -97,14 +105,16 @@ impl From<Position> for usize {
     }
 }
 
-///
+/// A set of leaves of a Merkle tree. 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum Leaf<A> {
     Left(A),
     Right(A, A),
 }
 
-///
+/// A `[NonEmptyFrontier]` is a reduced representation of a Merkle tree,
+/// having either one or two leaf values, and then a set of hashes produced
+/// by the reduction of previously appended leaf values. 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct NonEmptyFrontier<H> {
     position: Position,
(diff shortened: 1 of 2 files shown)


src/bridgetree.rs
```diff
@@ -8,43 +11,57 @@ use std::mem::size_of;
...
 
-    pub fn parent_levels(&self) -> impl Iterator<Item = Altitude> + '_ {
-        (0..=self.max_level().0).into_iter().filter_map(move |i| {
-            if i != 0 && self.0 & (1 << i) != 0 {
-                Some(Altitude(i))
-            } else {
-                None
-            }
-        })
+    /// Returns the altitude of each populated ommer
```
str4d, line 41:
```suggestion
    /// Returns the altitude of each populated ommer.
```

src/bridgetree.rs
```diff
@@ -8,43 +11,57 @@ use std::mem::size_of;
...
 
-    fn max_level(&self) -> Altitude {
+    /// Returns the altitude of the top of a binary tree containing
+    /// a number of nodes equal to the next power of two greater than 
+    /// or equal to `self + 1`.
+    fn max_altitude(&self) -> Altitude {
         Altitude(if self.0 == 0 {
             0
         } else {
             63 - self.0.leading_zeros() as u8
```
str4d, line 37: Separate from this PR: This looks like it will break on 32-bit platforms. This wouldn't affect `zcashd` (which only supports 64-bit targets), but could affect other Rust usages. We should explicitly cast to `u64` here.

src/bridgetree.rs
```diff
@@ -8,43 +11,57 @@ use std::mem::size_of;
...
+    /// path to the root of a merkle tree that has `self + 1` nodes.
+    pub fn altitudes_required_count(&self) -> usize {
+        self.altitudes_required().count()
     }
 
-    pub fn levels_required(&self) -> impl Iterator<Item = Altitude> + '_ {
-        (0..=(self.max_level() + 1).0)
+    /// Returns the altitude of each cousin and/or ommer required to construct 
+    /// an authentication path to the root of a merkle tree of depth `self + 1` 
+    /// nodes.
```
str4d, line 62:
```suggestion
    /// an authentication path to the root of a merkle tree that has `self + 1` 
    /// nodes.
```

src/bridgetree.rs
```diff
@@ -55,7 +72,10 @@ impl Position {
             })
     }
 
-    pub fn all_levels_required(&self) -> impl Iterator<Item = Altitude> + '_ {
+    /// Returns the altitude of each cousin and/or ommer required to construct 
+    /// an authentication path to the root of a merkle tree containing 2^64
+    /// nodes. 
+    pub fn all_altitudes_required(&self) -> impl Iterator<Item = Altitude> + '_ {
         (0..64).into_iter().filter_map(move |i| {
```
str4d, line 79: Here we also assume that `self.0` is 64 bits. Having the internal representation be `usize` is convenient for counting, but given there are several algorithms that assume a 64-bit representation, it would be safer to just have `Position(u64)`, and then use the `TryFrom<u64> for usize` anywhere we need to index something.

src/bridgetree.rs
```diff
@@ -86,107 +105,140 @@ impl From<Position> for usize {
...
+                .find(|(_, l)| *l == lvl)
+                .map(|(p, _)| p.clone())
+        }
+    }
+
+    /// Returns the value of the most recently appended leaf.
+    pub fn leaf_value(&self) -> H {
+        match &self.leaf {
+            Leaf::Left(v) => v.clone(),
+            Leaf::Right(_, v) => v.clone(),
```
str4d, line 156: Untested, but I think this works (and conveys a bit better that we are just performing value extraction):
```suggestion
            Leaf::Left(v) | Leaf::Right(_, v) => v.clone(),
```

str4d (commented):
Flushing comments.

src/bridgetree.rs
```diff
@@ -295,32 +306,40 @@ pub struct Frontier<H, const DEPTH: u8> {
...
         Frontier { frontier: None }
     }
 
-    /// Return the position of latest leaf appended to the frontier,
+    /// Constructs a new frontier from a `[NonEmptyFrontier]`
+    ///
+    /// Returns `None` if the provided frontier exceeds the maximum
+    /// allowed depth.
+    pub fn new(frontier: NonEmptyFrontier<H>) -> Option<Self> {
+        if frontier.size() >= 1 << DEPTH {
```
str4d, line 319: Double-check this edge case; I think this might actually need to be:
```suggestion
        if frontier.size() > 1 << DEPTH {
```

nuttycom, line 319:
```suggestion
        if frontier.size() > 1 << DEPTH {
```

src/bridgetree.rs
```diff
@@ -355,41 +374,51 @@ impl<H: Hashable + Clone, const DEPTH: u8> crate::Frontier<H> for Frontier<H, DE
     }
 }
 
+/// Each AuthFragment stores part of the authentication path for the leaf at a particular position.
+/// Successive fragments may be concatenated to produce the authentication path up to one less than
+/// the maximum altitude of the Merkle frontier corresponding to the leaf at the specified
+/// position. Then, the authentication path may be completed by hashing any
```
str4d, line 380:
```suggestion
/// position. Then, the authentication path may be completed by hashing any 
```
I want to know what happens in the next episode!

nuttycom, line 380:
```suggestion
/// position. Then, the authentication path may be completed by hashing with empty roots. 
```

src/bridgetree.rs
```diff
@@ -355,41 +388,51 @@ impl<H: Hashable + Clone, const DEPTH: u8> crate::Frontier<H> for Frontier<H, DE
...
+    /// the specified position separately from the length of the values vector because the values
+    /// will usually be split across multiple fragments.
+    altitudes_observed: usize,
+    /// The subtree roots at altitudes required for the position that have not been included in
+    /// preceding fragments.
     values: Vec<A>,
 }
 
 impl<A> AuthFragment<A> {
+    /// Construct the new empty authenticaiton path fragment for the specified position.
```
therealyingtong, line 395:
```suggestion
    /// Construct the new empty authentication path fragment for the specified position.
```

src/bridgetree.rs
```diff
@@ -355,41 +374,51 @@ impl<H: Hashable + Clone, const DEPTH: u8> crate::Frontier<H> for Frontier<H, DE
...
```
str4d, line 395:
```suggestion
    /// Construct the new empty authentication path fragment for the specified position.
```

src/lib.rs
```diff
@@ -31,11 +31,16 @@ use serde::{Deserialize, Serialize};
...
+/// A type-safe wrapper for indexing into "levels" of a binary tree, such that
+/// nodes at altitude `0` are leaves, nodes at altitude `1` are parents
+/// of nodes at altitude `0`, and so forth. This type is capable of 
+/// representing altitudes in trees containing up to 2^256 leaves.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(transparent)]
 pub struct Altitude(u8);
 
 impl Altitude {
+    /// Convenience method for returning the zero altitude
```
str4d, line 43:
```suggestion
    /// Convenience method for returning the zero altitude.
```

src/bridgetree.rs
nuttycom: Resolved in https://github.com/zcash/incrementalmerkletree/pull/6/commits/96342658e07747410a58557fe7e273b9e673d3f5

src/bridgetree.rs
nuttycom: Resolved in https://github.com/zcash/incrementalmerkletree/pull/6/commits/96342658e07747410a58557fe7e273b9e673d3f5

PR #6, opened 2021-07-08
Two minor fixes, plus methods required to support parsing and serialization.

nuttycom:
Two fixes are included here:
* Fix an off-by-one error in the computation of the maximum altitude of the frontier. This did not cause a divergence from correct authentication path computation because the maximum depth of the tree was checked separately from the generation of the list of required path components.
* Ensure that AuthFragment::fuse validates the compatibility of successive fragments being appended to one another. While in practice fragments should never be concatenated out of order, this was not previously being checked.

PR #7, opened 2021-07-16
Use witness for checkpointing

nuttycom:
This extends from #6

### DIFF
--- a/src/bridgetree.rs
+++ b/src/bridgetree.rs
@@ -118,7 +118,7 @@ impl<H: Hashable + Clone> NonEmptyFrontier<H> {
                 self.leaf = Leaf::Right(a.clone(), value);
             }
             Leaf::Right(a, b) => {
-                carry = Some((H::combine(Altitude::zero(), &a, &b), Altitude::one()));
+                carry = Some((H::combine(Altitude::zero(), a, b), Altitude::one()));
                 self.leaf = Leaf::Left(value);
             }
         };
@@ -128,7 +128,7 @@ impl<H: Hashable + Clone> NonEmptyFrontier<H> {
             for (ommer, ommer_lvl) in self.ommers.iter().zip(self.position.ommer_altitudes()) {
                 if let Some((carry_ommer, carry_lvl)) = carry.as_ref() {
                     if *carry_lvl == ommer_lvl {
-                        carry = Some((H::combine(ommer_lvl, &ommer, &carry_ommer), ommer_lvl + 1))
+                        carry = Some((H::combine(ommer_lvl, ommer, carry_ommer), ommer_lvl + 1))
                     } else {
                         // insert the carry at the first empty slot; then the rest of the
                         // ommers will remain unchanged
@@ -225,7 +225,7 @@ impl<H: Hashable + Clone> NonEmptyFrontier<H> {
 
             digest = H::combine(
                 ommer_lvl,
-                &ommer,
+                ommer,
                 // fold up to ommer.lvl pairing with empty roots; if
                 // complete_lvl == ommer.lvl this is just the complete
                 // digest to this point
@@ -560,7 +560,7 @@ impl<H: Hashable + Clone + PartialEq> MerkleBridge<H> {
     /// `self` had had every leaf used to construct `next` appended to it
     /// directly.
     fn fuse(&self, next: &Self) -> Option<MerkleBridge<H>> {
-        if next.can_follow(&self) {
+        if next.can_follow(self) {
             let fused = MerkleBridge {
                 prior_position: self.prior_position,
                 auth_fragments: self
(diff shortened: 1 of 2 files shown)


src/bridgetree.rs
```diff
@@ -227,100 +320,106 @@ impl<H: Hashable + Clone> NonEmptyFrontier<H> {
...
 impl<H, const DEPTH: u8> Frontier<H, DEPTH> {
-    /// Construct a new empty frontier.
-    pub fn new() -> Self {
+    /// Constructs a new empty frontier.
+    pub fn empty() -> Self {
         Frontier { frontier: None }
     }
 
-    /// Return the position of latest leaf appended to the frontier,
+    pub fn value(&self) -> Option<&NonEmptyFrontier<H>> {
```
daira, line 382: Should this not have a doc comment?

src/bridgetree.rs
```diff
@@ -355,50 +454,86 @@ impl<H: Hashable + Clone, const DEPTH: u8> crate::Frontier<H> for Frontier<H, DE
     }
 }
 
+/// Each AuthFragment stores part of the authentication path for the leaf at a particular position.
+/// Successive fragments may be concatenated to produce the authentication path up to one less than
+/// the maximum altitude of the Merkle frontier corresponding to the leaf at the specified
+/// position. Then, the authentication path may be completed by hashing any
```
daira, line 460: This comment seems to be incomplete.

src/bridgetree.rs
```diff
@@ -355,50 +454,86 @@ impl<H: Hashable + Clone, const DEPTH: u8> crate::Frontier<H> for Frontier<H, DE
...
         AuthFragment {
             position,
-            levels_observed: 0,
+            altitudes_observed: 0,
             values: vec![],
         }
     }
 
+    /// Construct a fragment from its component parts. This cannot
+    /// not perform any meaningful validation that the provided values
```
daira, line 485: Double negative?

src/lib.rs
```diff
@@ -31,11 +31,16 @@ use serde::{Deserialize, Serialize};
 use std::ops::Add;
 use std::ops::Sub;
 
+/// A type-safe wrapper for indexing into "levels" of a binary tree, such that
```
daira, line 34:
```suggestion
/// A type-safe wrapper for indexing into "altitudes" of a binary tree, such that
```

daira (approved):
utACK (modulo conflicts) with minor comments.

str4d:
Needs rebase now that #6 has merged.

src/bridgetree.rs
```diff
@@ -1373,6 +1395,19 @@ mod tests {
...
+    //    let mut t = BridgeTree::<String, 6>::new(100);
+    //    t.append(&"a".to_string());
+    //    t.checkpoint();
+    //    t.checkpoint();
+    //    assert!(t.rewind());
+    //    t.append(&"a".to_string());
+    //    assert!(t.rewind());
+    //    t.append(&"a".to_string());
+    //    assert_eq!(t.frontier().unwrap().root(), "aa______________");
+    //}
```
ebfull, line 1409: This test should be uncommented or removed.

ebfull (changes requested):

src/bridgetree.rs
```diff
@@ -922,32 +932,41 @@ impl<H: Hashable + Hash + Eq + Clone, const DEPTH: u8> Tree<H> for BridgeTree<H,
...
                 }
             }
-            Some(Checkpoint::AtIndex(i, bridge)) => {
-                // TODO: maybe there's a better way to do this check than
-                // searching the witnessed values twice?
-                if self.saved.values().any(|saved_idx| *saved_idx > i)
-                    || (self.saved.values().any(|saved_idx| *saved_idx == i)
-                        && bridge.frontier != self.bridges[i].frontier)
+            Some(Checkpoint::AtIndex(i)) => {
+                // Do not rewind if doing so would remove a witness.
```
str4d, line 936: This seems like incorrect behaviour to me. If we are rewinding to a checkpoint, that corresponds to "un-appending" from the incremental Merkle tree, which implies that whatever we were witnessing beyond that checkpoint no longer exists in the tree and cannot be witnessed.

src/bridgetree.rs
```diff
@@ -922,32 +932,41 @@ impl<H: Hashable + Hash + Eq + Clone, const DEPTH: u8> Tree<H> for BridgeTree<H,
...
+                        self.witness();
+                        true
+                    }
+                    _ => {
+                        // no witnesses at positions later than the checkpoint,
+                        // so we can just truncate.
+                        self.bridges.truncate(i + 1);
+                        // if the checkpoint removed was a duplicate, we need to
+                        // restore the successor bridge so that future appends correctly
+                        // affect the successor bridge
```
str4d, line 964: This is implicitly referring to the fact that rewinding to a checkpoint drops that checkpoint from the list, so someone that wants to rewind to the same block height multiple times (like `zcashd` during repeated reorgs) will need to remember to call `.checkpoint()` immediately after `.rewind()`.

Similarly, if I call `.checkpoint()`, append commitments from a block, call `.checkpoint()` again, and then immediately call `.rewind()` (because the block is immediately reorged away), I would expect to be rolled back to the first checkpoint, but the current code treats `.checkpoint().rewind()` as a no-op, so leave the `BridgeTree` in a state where we have one checkpoint, plus the commitments from the now-reorged block, but no checkpoint for that block.

src/bridgetree.rs
nuttycom: We confirmed that this was intended behavior in pairing with @ebfull.

src/bridgetree.rs
```diff
@@ -746,6 +761,40 @@ impl<H: Hashable + Hash + Eq + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+
+    fn witness_internal(&mut self, btype: BoundaryType) -> Option<usize> {
+        let next = self.bridges.last().map(|current| {
+            (
+                current.leaf_value().clone(),
+                current.successor(self.bridges.len() - 1),
+            )
+        });
+
+        next.map(|(leaf, succ)| {
```
daira, line 773: Wouldn't it be clearer to use ? here, e.g.:
```
        let current = self.bridges.last()?;
        let leaf = current.leaf_value().clone();
        let blen = self.bridges.len();
        let succ = current.successor(blen - 1);
```

nuttycom: Yeah, I guess that'd be the idiomatic Rust solution.

daira (approved):
utACK

src/bridgetree.rs
nuttycom: Oh, I remember why I did it this way now. It's that `self.bridges.last()?` immutably borrows `self.bridges` for the entire scope, and you then can't get the mutable borrow you need to push to it.

nuttycom: That's also the reason why I do this weird `let next = ...` and then `next.map` rather than doing everything in the scope of the single `map` call.

src/bridgetree.rs
```diff
@@ -683,6 +696,10 @@ impl<H: Hash + Eq, const DEPTH: u8> BridgeTree<H, DEPTH> {
     pub fn max_checkpoints(&self) -> usize {
         self.max_checkpoints
     }
+
+    pub fn frontier(&self) -> Option<&NonEmptyFrontier<H>> {
```
str4d, line 700: Document this. Non-blocking; there are a bunch of other APIs that also need documentation, so we should just have a separate PR for all of them.

str4d (approved):
utACK 0a8854f1b0885dd86cac502397e42b0f74b7d963

PR #14, opened 2022-02-22
Make tree rewinds fully restore the checkpointed state.

nuttycom:
Both the reference implementation and the bridgetree implementation previously exhibited a bug that could be demonstrated with the following trace of operations:

1. append x
2. witness
3. checkpoint
4. remove_witness x
5. rewind
6. auth_path x

Prior to this change, the `rewind` operation would leave the tree in a state where the authorization path could not be computed for `x`, because the witness removal was not reverted during the rewind operation. This PR gives `rewind` the semantics that it always reverts the tree and the set of witnesses being accumulated to exactly the state it had at the last checkpoint.


ebfull (approved):
ACK. My only contribution to this PR was the idea to change the way the sample tree worked so that checkpoints were stored as copies of the data structure, for robustness. So this can be merged now.

PR #15, opened 2022-02-22
Add garbage collection to the incremental merkle tree.

### LINKED ISSUE #2: Add mark_spent() method to Tree trait
There needs to be a way to mark a witness element as "spent" so that it is later garbage collected after a sufficient number of checkpoints are registered (or something along those lines) so that it can then be discarded, to avoid accumulating large numbers of witnesses. (Currently, remove_witness cannot be called unless we're rolling back the leaf being added to the tree.)

nuttycom:
This builds atop #14.

Fixes #2. Instead of a specific "mark for deferred removal" operation, all `remove_witness` calls simply mark a witness for garbage collection, which can be performed using the `BridgeTree::garbage_collect` method.


src/bridgetree.rs
```diff
@@ -758,6 +769,67 @@ impl<H: Hashable, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+        // a rewind.
+        if self.checkpoints.len() == self.max_checkpoints {
+            let gc_len = self.checkpoints.first().unwrap().bridges_len;
+            // Get a list of the leaf positions that we need to retain. This consists of
+            // all the saved leaves, plus all the leaves that have been forgotten since
+            // the most distant checkpoint to which we could rewind.
+            let remember: BTreeSet<H> = self
+                .saved
+                .keys()
+                .chain(self.checkpoints.iter().flat_map(|c| c.forgotten.keys()))
```
ebfull, line 785: I don't believe this will work properly if there are multiple leaves, is that being addressed in #16?

ebfull (approved):

src/bridgetree.rs
nuttycom: Yes.

PR #16, opened 2022-02-23
Use (position, hash) pair to uniquely identify leaves.

nuttycom:
Previously, we didn't correctly handle cases where the incremental merkle tree had duplicate leaf values; we just ignored every duplicate leaf added to the tree when constructing an authentication path. This PR allows for duplicate leaf values to be uniquely identified and correct authorization paths to be generated for all such leaves.

This builds atop #15

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,8 @@ pub mod bridgetree;
 mod sample;
 
 use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+use std::num::TryFromIntError;
 use std::ops::{Add, AddAssign, Sub};
 
 /// A type-safe wrapper for indexing into "levels" of a binary tree, such that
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -991,25 +992,90 @@ pub(crate) mod tests {
...
+    ) -> impl Strategy<Value = Vec<Operation<G::Value>>>
     where
         G::Value: Clone + 'static,
     {
-        item_gen.prop_flat_map(|item| {
+        let append_count = optypes
+            .iter()
+            .filter(|t| matches!(t, OpType::Append))
+            .count();
+        proptest::collection::vec(item_gen, append_count + 2)
```
ebfull, line 1053: Why is this `+ 2`?

ebfull (approved):

src/lib.rs
nuttycom: That should be + 1, I must have fat-fingered it. We want one more item than the number of appends so that we can always generate an `Authpath` or `Unwitness` even if we don't append any notes to the tree (those would always return false/None from the relevant operations).
```suggestion
        proptest::collection::vec(item_gen, append_count + 1)
```

PR #17, opened 2022-02-24
Return the current position & leaf from the witness operation.

nuttycom:


therealyingtong (approved):
utACK

src/bridgetree.rs
```diff
@@ -879,9 +879,10 @@ impl<H: Hashable, const DEPTH: u8> Tree<H> for BridgeTree<H, DEPTH> {
     }
 
     /// Marks the current tree state leaf as a value that we're interested in
-    /// witnessing. Returns true if successful and false if the tree is empty.
-    fn witness(&mut self) -> bool {
-        if let Some(key @ (position, _)) = self.current_leaf() {
+    /// witnessing. Returns the current position and leaf value if the treee
```
str4d, line 882:
```suggestion
    /// witnessing. Returns the current position and leaf value if the tree
```

src/lib.rs
```diff
@@ -245,11 +245,11 @@ pub trait Tree<H>: Frontier<H> {
...
-    /// Marks the current leaf as one for which we're interested in producing
-    /// an authentication path. Returns an optional value containing the current
-    /// position if successful or if the current value was already marked,
-    /// or None if the tree is empty.
-    fn witness(&mut self) -> bool;
+    /// Marks the current leaf as one for which we're interested in producing 
+    /// an authentication path. Returns an optional value containing the 
+    /// current position and leaf value if successful or if the current 
+    /// value was already marked, or None if the tree is empty.
+    fn witness(&mut self) -> Option<(Position, H)>;
```
str4d, line 252: I was going to ask for this to be documented in the changelog, but apparently we don't have one for this crate yet, so I'm not going to block on this. Open an issue for going through the API changes across crate releases and documenting them.

src/lib.rs
```diff
@@ -245,11 +245,11 @@ pub trait Tree<H>: Frontier<H> {
     /// the specified leaf value from the specified position in the tree.
     fn is_witnessed(&self, position: Position, value: &H) -> bool;
 
-    /// Marks the current leaf as one for which we're interested in producing
-    /// an authentication path. Returns an optional value containing the current
-    /// position if successful or if the current value was already marked,
-    /// or None if the tree is empty.
-    fn witness(&mut self) -> bool;
+    /// Marks the current leaf as one for which we're interested in producing
```
str4d, line 248:
```suggestion
    /// Marks the current leaf as one for which we're interested in producing
```
I also don't understand why the documentation change for this function is present: it seems like the documentation was incorrect before but correct now?

src/lib.rs
```diff
@@ -782,8 +782,9 @@ pub(crate) mod tests {
         }
 
         /// Marks the current tree state leaf as a value that we're interested in
-        /// witnessing. Returns true if successful and false if the tree is empty.
-        fn witness(&mut self) -> bool {
+        /// witnessing. Returns the current position and leaf value if the treee
```
str4d, line 785:
```suggestion
        /// witnessing. Returns the current position and leaf value if the tree
```

src/sample.rs
```diff
@@ -65,17 +65,15 @@ impl<H: Hashable + PartialEq + Clone> TreeState<H> {
     }
 
     /// Marks the current tree state leaf as a value that we're interested in
-    /// witnessing. Returns true if successful and false if the tree is empty.
-    fn witness(&mut self) -> bool {
-        if let Some((pos, value)) = self.current_leaf() {
+    /// witnessing. Returns the current position and leaf value if the treee
```
str4d, line 68:
```suggestion
    /// witnessing. Returns the current position and leaf value if the tree
```

src/sample.rs
```diff
@@ -207,8 +205,9 @@ impl<H: Hashable + PartialEq + Clone> Tree<H> for CompleteTree<H> {
     }
 
     /// Marks the current tree state leaf as a value that we're interested in
-    /// witnessing. Returns true if successful and false if the tree is empty.
-    fn witness(&mut self) -> bool {
+    /// witnessing. Returns the current position and leaf value if the treee
```
str4d, line 208:
```suggestion
    /// witnessing. Returns the current position and leaf value if the tree
```

str4d (changes requested):

str4d (approved):
utACK after above documentation fixes.

PR #18, opened 2022-02-24
Fix formatting & linting errors.

nuttycom:
(and test CI)

### DIFF
--- a/.github/workflows/lints-beta.yml
+++ b/.github/workflows/lints-beta.yml
@@ -1,11 +1,10 @@
-name: Nightly lints
+name: Beta lints
 
-# These lints are only informative, so we only run them directly on branches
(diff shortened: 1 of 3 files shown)


steven-ecc (approved):
utACK

.github/workflows/lints-beta.yml
```diff
@@ -1,11 +1,10 @@
-name: Nightly lints
+name: Beta lints
 
-# These lints are only informative, so we only run them directly on branches
-# and not trial-merges of PRs, to reduce noise.
-on: push
+# We only run these lints on trial-merges of PRs to reduce noise.
+on: pull_request
```
str4d, line 4: I guess it doesn't matter which one we choose.

PR #20, opened 2022-02-24
Do not force Hashable values to also be Ord and Clone

nuttycom:
While these additional bounds are often used with hashable
types, they are not necessary for the construction of `Frontier`
values, and so should not be required everywhere.


str4d (approved):
utACK

PR #24, opened 2022-03-11
Fix incorrect consistency check for auth_fragment construction.

nuttycom:
This was just entirely wrong; it will often be the case that we've
observed several altitudes but have an empty list of values within
a particular fragment.

### DIFF
--- a/src/bridgetree.rs
+++ b/src/bridgetree.rs
@@ -374,7 +374,6 @@ impl<A> AuthFragment<A> {
     /// perform any meaningful validation that the provided values
     /// are valid.
     pub fn from_parts(position: Position, altitudes_observed: usize, values: Vec<A>) -> Self {
-        assert!(altitudes_observed <= values.len());
         Self {
             position,
             altitudes_observed,


steven-ecc (approved):
utACK

src/bridgetree.rs
```diff
@@ -374,7 +374,6 @@ impl<A> AuthFragment<A> {
     /// perform any meaningful validation that the provided values
     /// are valid.
     pub fn from_parts(position: Position, altitudes_observed: usize, values: Vec<A>) -> Self {
-        assert!(altitudes_observed <= values.len());
```
str4d, line 377: I am satisfied that the removal of this assertion between fields is consistent with their documentation:
```rust
    /// We track the total number of altitudes collected across all fragments
    /// constructed for the specified position separately from the length of
    /// the values vector because the values will usually be split across multiple
    /// fragments.
    altitudes_observed: usize,
    /// The subtree roots at altitudes required for the position that have not
    /// been included in preceding fragments.
    values: Vec<A>,
```
`altitudes_observed` covers all fragments, and notes that the collected values will usually be split across multiple fragments; `values` is the subtree roots not included in previous fragments. It makes sense that it's possible for `altitudes_observed > values.len()`, or indeed `altitudes_observed > 0 && values.is_empty()`.

str4d (approved):
utACK

daira (commented):
Post-hoc ACK

PR #29, opened 2022-03-30
Fix a bug in garbage collection.

nuttycom:


nuttycom:
#31 is a much better solution.

PR #31, opened 2022-03-30
Track the mutable tip of the bridgetree separate from the prior bridges.

nuttycom:
This simplifies a lot of the reasoning about checkpoints and rewinds, making
it no longer necessary to do a lot of index math.

### DIFF
--- a/src/bridgetree.rs
+++ b/src/bridgetree.rs
@@ -839,13 +839,13 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
                 .cloned()
                 .collect();
 
-            let mut pruned_bridges: Vec<MerkleBridge<H>> = vec![];
             let mut cur: Option<MerkleBridge<H>> = None;
             let mut merged = 0;
             let mut prune_fragment_positions: BTreeSet<Position> = BTreeSet::new();
-            // TODO: I really want to use `into_iter` here, but can't because self.bridges is
-            // behind a mut reference?
-            for (i, next_bridge) in self.prior_bridges.iter().enumerate() {
+            for (i, next_bridge) in std::mem::take(&mut self.prior_bridges)
+                .into_iter()
+                .enumerate()
+            {
                 if let Some(cur_bridge) = cur {
                     let witness_key = cur_bridge.tip();
                     let mut new_cur = if remember.contains(&witness_key) || i > gc_len {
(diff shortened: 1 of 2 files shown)


src/bridgetree.rs
```diff
@@ -725,9 +732,14 @@ impl<H: Ord, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
 
-    /// Returns the bridges that make up this tree
-    pub fn bridges(&self) -> &[MerkleBridge<H>] {
-        &self.bridges
+    /// Returns the prior bridges that make up this tree
+    pub fn prior_bridges(&self) -> &[MerkleBridge<H>] {
+        &self.prior_bridges
+    }
+
+    /// Returns the current bridges at the tip of this tree
```
daira, line 740:
```suggestion
    /// Returns the current bridge at the tip of this tree
```

daira (approved):
utACK, but I did not check the logic changes in detail.

src/bridgetree.rs
```diff
@@ -747,50 +759,64 @@ impl<H: Ord, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+            .zip(prior_bridges.iter().skip(1))
             .any(|(prev, next)| !next.can_follow(prev))
         {
             return Err(BridgeTreeError::ContinuityError);
         }
 
+        if prior_bridges
+            .last()
+            .zip(current_bridge.as_ref())
+            .map_or(false, |(prev, next)| !next.can_follow(prev))
```
ebfull, line 812: Wouldn't this not work properly (aka return a continuity error) if `prior_bridges` was empty, which should be a valid situation?

nuttycom: Good point, this should probably be `map_or(true,`...

src/bridgetree.rs
```diff
@@ -1242,4 +1301,18 @@ mod tests {
...
+        let mut tree: BridgeTree<String, 7> = BridgeTree::new(100);
+        let empty_root = tree.root();
+        tree.append(&"a".to_string());
+        for _ in 0..100 {
+            tree.checkpoint();
+        }
+        tree.garbage_collect();
+        assert!(tree.root() != empty_root);
+        tree.rewind();
+        assert!(tree.root() != empty_root); // this panics
```
ebfull, line 1316: Remove this comment.

ebfull (approved):

PR #38, opened 2022-04-09
Add as_of_root argument to Tree::authentication_path

nuttycom:
This allows computation of the authentication path as of previous tree states, in
addition to the previous behavior which only allowed computation of the path as of the
most recent tree state. The provided `as_of_root` value must be equal to either the
current root of the tree, or to the root of the tree at a previous checkpoint.

Best reviewed by commit.


CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to Rust's notion of
...
 ## [Unreleased]
 
+### Changed
+
+- `incrementalmerkletree`:
+  - `Tree::authentication_path` has been changed to take an `as_of_root` parameter.
+    This allows computation of the authentication path as of previous tree states, in
+    addition to the previous behavior which only allowed computation of the path as of the
+    most recent tree state. The provided `as_of_root` value must be equal to either the
+    current root of the tree, or to the root of the tree at a previous checkpoint.
```
daira, line 17: `Checkpoint` is public, so the addition of the `witnessed` field and corresponding method changes need to be documented.

src/bridgetree.rs
```diff
@@ -1007,24 +1068,92 @@ impl<H: Hashable + Ord + Clone + Debug, const DEPTH: u8> Tree<H> for BridgeTree<
...
+        // Find the earliest checkpoint having a matching root, or the current
+        // root if it matches and there is no earlier matching checkpoint.
+        let auth_base = self
+            .checkpoints
+            .iter()
+            .enumerate()
+            .rev()
+            .take_while(|(_, c)| c.position(&self.prior_bridges) >= Some(position))
+            .filter(|(_, c)| &c.root(&self.prior_bridges, max_alt) == as_of_root)
+            .last()
```
daira, line 1090: `.rev().find(...)` might be a little more efficient than `.filter(...).last()`.

src/bridgetree.rs
```diff
@@ -1007,24 +1068,92 @@ impl<H: Hashable + Ord + Clone + Debug, const DEPTH: u8> Tree<H> for BridgeTree<
...
+                    AuthBase::Current
+                } else {
+                    AuthBase::NotFound
+                }
+            });
+
+        let saved_idx = self.saved.get(&position).or_else(|| {
+            if let AuthBase::Checkpoint(i, _) = auth_base {
+                // The saved position might have been forgotten since the checkpoint,
+                // so look for it in each of the subsequent checkpoints forgotten
```
daira, line 1103:
```suggestion
               // so look for it in each of the subsequent checkpoints' forgotten
```

src/bridgetree.rs
nuttycom: That seems unlikely to me, because assuming that we use 10 as the normal checkpoint depth, this will ordinarily require 10 evaluations rather than 90.

nuttycom: Oh, wait, I see what you're saying. I guess that if `rev` is just manipulating indexed lookups, then this could save a few evaluations of the filter.

nuttycom: I tried it out and it doesn't work anyway; `take_while`'s result type doesn't implement `DoubleEndedIterator`, which is what you need for the `reverse`.

ebfull (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,29 @@ and this project adheres to Rust's notion of
...
+    `BridgeTree::witness` while this checkpoint was the current checkpoint.
+
+### Changed
+
+- `incrementalmerkletree`:
+  - `Tree::authentication_path` has been changed to take an `as_of_root` parameter.
+    This allows computation of the authentication path as of previous tree states, in
+    addition to the previous behavior which only allowed computation of the path as of the
+    most recent tree state. The provided `as_of_root` value must be equal to either the
+    current root of the tree, or to the root of the tree at a previous checkpoint.
```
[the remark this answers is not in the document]
nuttycom:
```suggestion
    current root of the tree, or to the root of the tree at a previous checkpoint that
    contained a note at the given position. 
```

daira (approved):
utACK

PR #42, opened 2022-07-20
Alter MerkleBridge to avoid storing duplicate ommers.

nuttycom:
Prior to this change, each position being tracked would
require an entry in the `auth_fragments` map for each
bridge in a BridgeTree. However, very frequently, these
entries would duplicate storage of nodes higher up in the
tree, wherever a single node in the ??missingword?? was an ancestor of
a leaf being tracked.

This commit restructures the tracking of auth data to
avoid this duplication of data, as well as the redundant
work that was required to produce each copy of the duplicated
nodes.

### DIFF
--- a/src/bridgetree.rs
+++ b/src/bridgetree.rs
@@ -8,8 +8,9 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::mem::size_of;
+use std::ops::Range;
 
-use super::{Altitude, Hashable, Position, Tree};
+use super::{Address, Altitude, Hashable, Position, Source, Tree};
 
 /// A set of leaves of a Merkle tree.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


daira:
There's a missing word in the description. I wasn't sure what it should be.

src/bridgetree.rs
```diff
@@ -527,34 +489,51 @@ impl<'a, H: Hashable + Ord + Clone + 'a> MerkleBridge<H> {
...
 
-        for ext in self.auth_fragments.values_mut() {
-            ext.augment(&self.frontier);
+        let mut found = vec![];
+        for address in self.tracking.iter() {
+            // We know that there will only ever be one address that we're
+            // tracking at a given level, because as soon as we find a
+            // value for the sibling of the address we're tracking, we
+            // remove the tracked address and replace it the next parent
+            // of that address for which we need to find a sibling.
```
daira, line 522: Ohh 💡 :-)

daira (approved):
utACK

ebfull (approved):

PR #43, opened 2022-07-22
    Remove the `bridgetree::Leaf` type & make ommer levels start at 0

nuttycom:
This is a major refactoring that gets rid of a significant wart
in the design of the `bridgetree::Frontier` type that has existed
since the the project's inception.

The `Leaf` type incorrectly treated the siblings of leaves at
odd positions as special, whereas they should have been simply
treated as ordinary entries in the frontier's ommers list. Correcting
this allows for the code to be significantly simplified in a number
of places.

This builds atop #42


ebfull (approved):
ACK, spent some time looking at this with @nuttycom as well. Much more sensible design.

PR #44, opened 2022-07-22
Rework the API for move to the `bridgetree` crate.

nuttycom:
This builds atop #43. It clarifies the naming of a number of operations, and moves the `Frontier` and `Tree` traits to a `testing` module while moving the `BridgeTree` type to the top level.


src/lib.rs
```diff
@@ -1,1556 +1,1510 @@
...
-        (self.0..other.0).into_iter().map(Level)
-    }
+use std::collections::{BTreeMap, BTreeSet};
+use std::convert::TryFrom;
+use std::fmt::Debug;
+use std::mem::size_of;
+use std::ops::Range;
+
+use crate::{
+    hashing::Hashable,
```
ebfull, line 47: We should `pub use` this trait at the `lib.rs` level, and not expose the `hashing` module publicly.

src/lib.rs
```diff
@@ -1,1556 +1,1510 @@
...
+    PositionNotMarked(Position),
+    BridgeFusionError(ContinuityError),
+    BridgeAddressInvalid(Address),
 }
 
-impl From<Level> for u8 {
-    fn from(level: Level) -> u8 {
-        level.0
-    }
+/// A `[NonEmptyFrontier]` is a reduced representation of a Merkle tree,
```
ebfull, line 86:
```suggestion
/// A [`NonEmptyFrontier`] is a reduced representation of a Merkle tree,
```

src/lib.rs
```diff
@@ -1,1556 +1,1510 @@
...
     }
 }
 
-impl From<Position> for Address {
-    fn from(p: Position) -> Self {
-        Address {
-            level: 0.into(),
-            index: p.into(),
+impl<'a, H: Hashable + Ord + Clone + 'a> MerkleBridge<H> {
+    /// Constructs a new bridge to follow this one. If mark_current_leaf is true, the successor
```
ebfull, line 425:
```suggestion
    /// Constructs a new bridge to follow this one. If `mark_current_leaf` is true, the successor
```

src/lib.rs
```diff
@@ -1,1556 +1,1510 @@
...
+}
 
-    /// Remove state from the tree that no longer needs to be maintained
-    /// because it is associated with checkpoints or witnesses that
-    /// have been removed from the tree at positions deeper than those
-    /// reachable by calls to `rewind`. It is always safe to implement
-    /// this as a no-op operation
-    fn garbage_collect(&mut self);
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct BridgeTree<H, const DEPTH: u8> {
```
ebfull, line 661: This structure isn't documented -- probably should be at least a little considering it's the whole point of the crate. :P

src/lib.rs
```diff
@@ -1,1556 +1,1510 @@
...
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub(crate) enum Source {
-    /// The sibling to the address can be derived from the incremental frontier
-    /// at the contained ommer index
-    Past(usize),
-    /// The sibling to the address must be obtained from values discovered by
-    /// the addition of more nodes to the tree
-    Future,
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MerkleBridge<H> {
```
ebfull, line 314: This structure is also not documented.

ebfull (approved):
ACK modulo (mostly) nits

src/lib.rs
nuttycom: Hah, yeah, I noticed this when using the crate in librustzcash.

ebfull (approved):

ebfull (commented):
post-hoc ACK

PR #45, opened 2022-08-04
Add conversions from incrementalmerkletree types.

nuttycom:
A new feature flag, `incrementalmerkletree-compat` is introduced.
This flag guards a set of conversions from `incrementalmerkletree`
types to `bridgetree` types to facilitate updating to `bridgetree`
from `incrementalmerkletree`.


nuttycom:
Obsolete.

PR #47, opened 2022-08-16
Make it possible to tag checkpoints with identifying information.

nuttycom:
When restoring bridgetree information from persistent data, it
is often useful and/or necessary to be able to validate the state
of the returned data against other information in the system that
is using the tree. An easy way to do this is to add identifying
information to each checkpoint so that it can be compared to the
state of the system that might otherwise need to make dangerous
assumptions about the state of the tree as of a given checkpoint.


daira (approved):
Non-blocking: please add a test for the ID functionality, and a changelog entry. Otherwise utACK.

PR #50, opened 2022-12-15
Add `shardtree` crate.

### LINKED ISSUE #48: Implement a sparse Merkle tree type to allow out-of-order insertion.

nuttycom:
This builds atop #52 

The `shardtree` crate implements a sparse, aggressively pruned Merkle tree that supports out-of-order insertion, witnessing of marked leaves, and truncation.

Closes #48.


nuttycom:
Converting back to draft as I figured out a simpler way to approach truncation in root & witness calculation.

softminus:
It feels to me like there's a bit too much freedom the way the `PrunableTree` is defined:
```
type PrunableTree<H> = Tree<Option<Rc<H>>, (H, RetentionFlags)>;
```

since this means that leaf nodes (if the `Option` isn't a `None`) can have two (potentially different) values of type `H` (here, hashes), and this seems like it would be a significant pain to ensure that the obvious invariant of "the two hashes are always the same" always holds regardless of the tree operations.

Is there a way to use the type system to constrain `Leaf` nodes to only have one single value of type `H`?

softminus (changes requested):

nuttycom:
[force-pushed](https://github.com/zcash/incrementalmerkletree/compare/2494c3008687e5366019422eaaf2f55ff65ebc30..3b63465b3e42534b4da9888bf35af35ed7bd7d60) to prune the annotation state space to eliminate some invalid states.

shardtree/src/lib.rs
```diff
@@ -151,11 +255,267 @@ impl<A: Clone, V: Clone> Tree<A, V> {
...
+
+    /// Returns the Merkle root of this tree, given the address of the root node.
+    ///
+    /// Along with the root of the Merkle tree, this returns the minimum position at which a `Nil`
+    /// leaf was replaced with a value provided by [`H::empty_root()`]. If no empty roots were
+    /// introduced, then this value will be equal to `root_addr.position_range_end()`.
+    ///
+    /// ### Parameters:
+    /// * `truncate_to` An inclusive lower bound on positions in the tree beyond which all
+    ///   leaf values will be treated as `Nil`.
```
ebfull, line 285: I think this doc comment is out of date. The parameter for this function is called `truncate_at`, and it doesn't sound like the return type matches the description.

shardtree/src/lib.rs
```diff
@@ -151,11 +255,267 @@ impl<A: Clone, V: Clone> Tree<A, V> {
...
+                // While we could handle that here, if we encountered that case it would
+                // be indicative of a programming error elsewhere and so we assert instead.
+                assert!(!(left.0.is_nil() && right.0.is_nil()));
+                let (left_root, right_root) = root_addr
+                    .children()
+                    .expect("A parent node cannot appear at level 0");
+
+                let mut left_incomplete = left.marked_positions(left_root);
+                let mut right_incomplete = right.marked_positions(right_root);
+                left_incomplete.append(&mut right_incomplete);
```
ebfull, line 342: The slightly more "correct" thing to do here is `left_incomplete.extend(right_incomplete);` which takes `right_incomplete` by value and doesn't require `right_incomplete` to be mutable.

shardtree/src/lib.rs
```diff
@@ -151,11 +255,267 @@ impl<A: Clone, V: Clone> Tree<A, V> {
...
+// accumulate errors, but we don't have one so we just write a special- cased version here.
+fn accumulate_result_with<A, B, C>(
+    left: Result<A, Vec<Address>>,
+    right: Result<B, Vec<Address>>,
+    combine_success: impl FnOnce(A, B) -> C,
+) -> Result<C, Vec<Address>> {
+    match (left, right) {
+        (Ok(a), Ok(b)) => Ok(combine_success(a, b)),
+        (Err(mut xs), Err(mut ys)) => {
+            xs.append(&mut ys);
```
ebfull, line 2208: `xs.extend(ys);` and `ys` does not need to be mutable.

ebfull (changes requested):
First thorough review pass of the first several commits. It seems like a lot of stuff is ending up in the public API that I would think didn't need to be there ultimately, and it would perhaps be nicer to break some of the structures into separate submodules especially to help identify those cases.

shardtree/src/lib.rs
ebfull: Actually, I'm not sure that is actually a more correct way given Rust doesn't have specialization. It might actually be slower because `append` on `BTreeSet` can avoid iterating. Nevermind!

softminus (changes requested):

softminus (changes requested):

shardtree/src/lib.rs
```diff
@@ -141,11 +245,281 @@ impl<A, V: Clone> Tree<A, V> {
...
+                        Err(vec![root_addr])
+                    }
+                }
+                Tree(ann, Node::Nil) => {
+                    if truncate_at >= root_addr.position_range_end() {
+                        // we can use the annotated root here, if any exists
+                        ann.as_ref()
+                            .map(|rc| rc.as_ref().clone())
+                            .ok_or_else(|| vec![root_addr])
+                    } else {
```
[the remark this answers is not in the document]
nuttycom: This case can no longer occur.

shardtree/src/lib.rs
```diff
@@ -141,11 +245,281 @@ impl<A, V: Clone> Tree<A, V> {
...
+        match (left, right) {
+            (Tree(_, Node::Nil), Tree(_, Node::Nil)) => Tree(ann, Node::Nil),
+            (Tree(_, Node::Leaf { value: lv }), Tree(_, Node::Leaf { value: rv }))
+                // we can prune right-hand leaves that are not marked; if a leaf
+                // is a checkpoint then that information will be propagated to
+                // the replacement leaf
+                if lv.1 == EPHEMERAL && (rv.1 & MARKED) == EPHEMERAL =>
+            {
+                Tree(
+                    ann,
```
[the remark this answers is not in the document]
nuttycom: This possibility is no longer representable.

shardtree/src/lib.rs
```diff
@@ -1,9 +1,94 @@
...
+/// of the witness for a leaf with `CHECKPOINT` or `MARKED` retention.
+pub static EPHEMERAL: RetentionFlags = RetentionFlags(0b00000000);
+
+/// A leaf with `CHECKPOINT` retention can be pruned when there are more than `max_checkpoints`
+/// additional checkpoint leaves, if it is not also a marked leaf.
+pub static CHECKPOINT: RetentionFlags = RetentionFlags(0b00000001);
+
+/// A leaf with `MARKED` retention can be pruned only as a consequence of an explicit deletion
+/// action.
+pub static MARKED: RetentionFlags = RetentionFlags(0b00000010);
```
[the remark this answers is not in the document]
nuttycom: `bitflags` looks nice, I didn't know that existed.

nuttycom:
[force-pushed](https://github.com/zcash/incrementalmerkletree/compare/d604ffc14a1ea2a8e4de2a900b12f6fdcb9de33a..fa7c6673fd18287789925f651cdeea473a7b2193) to address comments from review.

nuttycom:
> First thorough review pass of the first several commits. It seems like a lot of stuff is ending up in the public API that I would think didn't need to be there ultimately, and it would perhaps be nicer to break some of the structures into separate submodules especially to help identify those cases.

My general development philosophy, particularly when immutable data structures are concerned, is that one should make public all the operations that are possible to make public. If you have specific suggestions about things that you think are unsafe to have in the public API though, please make note of them.

shardtree/src/lib.rs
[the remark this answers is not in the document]
nuttycom: Do you think it's actually worth it to bring in the extra dependency? One of the things that I like about doing this manually is that the flags and the flag masks can have different types, with the relationships between those types correctly represented in the type system.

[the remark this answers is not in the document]
nuttycom: I suppose that I could still create the separate mask types, but really the existing operations are all that we need.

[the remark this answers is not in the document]
ebfull: I do think it's worth it because we already depend on it in our Rust tech stack in Zcash and it's a pretty common dependency in the Rust ecosystem.

shardtree/src/lib.rs
```diff
@@ -1 +1,242 @@
...
+                let left_result = left.reduce(alg);
+                let right_result = right.reduce(alg);
+                alg(Node::Parent {
+                    ann: ann.clone(),
+                    left: left_result,
+                    right: right_result,
+                })
+            }
+            Node::Leaf { value } => alg(Node::Leaf {
+                value: value.clone(),
```
ebfull, line 240: I think it'd be better for the reduction function to take `F: for<'a> Fn(Node<B, &'a A, &'a V>) -> B` instead so that you don't need to perform this clone or the clone of the annotation above, improving performance and reducing type restrictions.

shardtree/src/lib.rs
```diff
@@ -1 +1,242 @@
...
+                right.try_reduce(alg).right_and_then(move |r_value| {
+                    alg(Node::Parent {
+                        ann: ann.clone(),
+                        left: l_value,
+                        right: r_value,
+                    })
+                })
+            }),
+            Node::Leaf { value } => alg(Node::Leaf {
+                value: value.clone(),
```
ebfull, line 263: Same comment on the clone above.

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2851 @@
...
+            addr: Address,
+            t0: PrunableTree<H>,
+            t1: PrunableTree<H>,
+        ) -> Result<PrunableTree<H>, Address> {
+            // Require that any roots the we compute will not be default-filled by picking
+            // a starting valid fill point that is outside the range of leaf positions.
+            let no_default_fill = addr.position_range_end();
+            match (t0, t1) {
+                (Tree(Node::Nil), other) => Ok(other),
+                (other, Tree(Node::Nil)) => Ok(other),
```
ebfull, line 413: These can be merged together.

```suggestion
            (Tree(Node::Nil), other) | (other, Tree(Node::Nil)) => Ok(other),
```

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2851 @@
...
+                    if parent
+                        .root_hash(addr, no_default_fill)
+                        .iter()
+                        .all(|r| r == &value.0)
+                    {
+                        Ok(parent.reannotate_root(Some(Rc::new(value.0))))
+                    } else {
+                        Err(addr)
+                    }
+                }
```
ebfull, line 444: These two cases can be merged together because they behave the same. Also, we can make the pattern match more exhaustive (and the code within more robust) by explicitly pattern matching on the `Node::Parent` case. (The Rust compiler doesn't actually "know" or remember the pattern match holds within the case body, which is annoying.)

```suggestion
               (Tree(Node::Leaf { value }), parent @ Tree(Node::Parent { .. }))
                | (parent @ Tree(Node::Parent { .. }), Tree(Node::Leaf { value })) => {
                    if parent
                        .root_hash(addr, no_default_fill)
                        .iter()
                        .all(|r| r == &value.0)
                    {
                        Ok(parent.reannotate_root(Some(Rc::new(value.0))))
                    } else {
                        Err(addr)
                    }
                }
```

shardtree/src/lib.rs
```diff
@@ -1436,6 +1438,383 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+    store: S,
+    /// The maximum number of checkpoints to retain before pruning.
+    max_checkpoints: usize,
+    /// A map from position to the count of checkpoints at this position.
+    checkpoints: BTreeMap<C, Checkpoint>,
+    // /// A tree that is used to cache the known roots of subtrees in the "cap" of nodes between
+    // /// `SHARD_HEIGHT` and `DEPTH` that are otherwise not directly represented in the tree.  This
+    // /// cache is automatically updated when computing roots and witnesses. Leaf nodes are empty
+    // /// because the annotation slot is consistently used to store the subtree hashes at each node.
+    // cap_cache: Tree<Option<Rc<H>>, ()>
```
ebfull, line 1566: Is this intended to be removed?

nuttycom: It's more of a TODO to be implemented later.

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2851 @@
...
+
+        let mut result = BTreeSet::new();
+        go(self.root_addr, &self.root, &mut result);
+        result
+    }
+
+    /// Compute the witness for the leaf at the specified position.
+    ///
+    /// This tree will be truncated to the `truncate_at` position, and then empty
+    /// empty roots corresponding to later positions will be filled by [`H::empty_root`].
```
nuttycom, line 821:
```suggestion
    /// This tree will be truncated to the `truncate_at` position, and then empty
    /// roots corresponding to later positions will be filled by [`H::empty_root`].
```

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2851 @@
...
+                                left.leaf_value()
+                                    .map(|v| vec![v.clone()])
+                                    .ok_or_else(|| vec![l_addr])
+                            } else {
+                                Err(vec![l_addr])
+                            }
+                        } else if left.is_marked_leaf() {
+                            // If we have the left-hand leaf and the right-hand leaf is empty, we
+                            // can fill it with the empty leaf, but only if `fill_start` is None or
+                            // it is located at `position + 1`.
```
nuttycom, line 875:
```suggestion
                            // can fill it with the empty leaf, but only if we are truncating at
                            // a position to the left of the current position
```

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2851 @@
...
+    store: S,
+    /// The maximum number of checkpoints to retain before pruning.
+    max_checkpoints: usize,
+    /// A map from position to the count of checkpoints at this position.
+    checkpoints: BTreeMap<C, Checkpoint>,
+    // /// A tree that is used to cache the known roots of subtrees in the "cap" of nodes between
+    // /// `SHARD_HEIGHT` and `DEPTH` that are otherwise not directly represented in the tree.  This
+    // /// cache is automatically updated when computing roots and witnesses. Leaf nodes are empty
+    // /// because the annotation slot is consistently used to store the subtree hashes at each node.
+    // cap_cache: Tree<Option<Rc<H>>, ()>
```
nuttycom, line 1566:
```suggestion
    // /// TODO: Add a tree that is used to cache the known roots of subtrees in the "cap" of nodes between
    // /// `SHARD_HEIGHT` and `DEPTH` that are otherwise not directly represented in the tree.  This
    // /// cache will be automatically updated when computing roots and witnesses. Leaf nodes are empty
    // /// because the annotation slot is consistently used to store the subtree hashes at each node.
    // cap_cache: Tree<Option<Rc<H>>, ()>
```

shardtree/src/lib.rs
```diff
@@ -1436,6 +1438,383 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+        self.iter().map(|s| s.root_addr).collect()
+    }
+
+    fn truncate(&mut self, from: Address) -> Result<(), Self::Error> {
+        self.truncate(from.index());
+        Ok(())
+    }
+}
+
+/// A left-dense, sparse binary Merkle tree of the specified depth, represented as a vector of
```
softminus, line 1457: > left-dense

possibly remove this?

softminus: also "vector"

shardtree/src/lib.rs
```diff
@@ -1436,6 +1438,383 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+/// This tree maintains a collection of "checkpoints" which represent positions, usually near the
+/// front of the tree, that are maintained such that it's possible to truncate nodes to the right
+/// of the specified position.
+#[derive(Debug)]
+pub struct ShardTree<H, C: Ord, S: ShardStore<H>, const DEPTH: u8, const SHARD_HEIGHT: u8> {
+    /// The vector of tree shards.
+    store: S,
+    /// The maximum number of checkpoints to retain before pruning.
+    max_checkpoints: usize,
+    /// A map from position to the count of checkpoints at this position.
```
softminus, line 1469: correct comment re what maps to what

shardtree/src/lib.rs
```diff
@@ -1436,6 +1438,383 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+    // /// `SHARD_HEIGHT` and `DEPTH` that are otherwise not directly represented in the tree.  This
+    // /// cache is automatically updated when computing roots and witnesses. Leaf nodes are empty
+    // /// because the annotation slot is consistently used to store the subtree hashes at each node.
+    // cap_cache: Tree<Option<Rc<H>>, ()>
+    _hash_type: PhantomData<H>,
+}
+
+impl<
+        H: Hashable + Clone + PartialEq,
+        C: Clone + Ord + core::fmt::Debug,
```
softminus, line 1481: `Debug` isn't necessary here

shardtree/src/lib.rs
```diff
@@ -1436,6 +1438,383 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+            let mut clear_positions: BTreeMap<Address, BTreeMap<Position, RetentionFlags>> =
+                BTreeMap::new();
+            for (cid, checkpoint) in self
+                .checkpoints
+                .iter()
+                .take(self.checkpoints.len() - self.max_checkpoints)
+            {
+                checkpoints_to_delete.push(cid.clone());
+
+                // clear the checkpoint leaf
```
softminus, line 1814: maybe factor out this and the following for loop

shardtree/src/lib.rs
```diff
@@ -1713,6 +1713,98 @@ impl<
...
+                    let (l_addr, r_addr) = root_addr.children().unwrap();
+                    go(r_addr, right).map_or_else(
+                        || {
+                            go(l_addr, left).map(|(new_left, pos)| {
+                                (
+                                    Tree::unite(
+                                        l_addr.level(),
+                                        ann.clone(),
+                                        new_left,
+                                        right.as_ref().clone(),
```
softminus, line 1725: does the semantics of the `map_or_else` imply that `right` is going to be a Nil node? Should we add an assert here or something? Do we need to use `unite` at all?

softminus:
utACK given review comments

softminus (approved):
utACK pending changes

shardtree/src/lib.rs
```diff
@@ -0,0 +1,2763 @@
...
+    store: S,
+    /// The maximum number of checkpoints to retain before pruning.
+    max_checkpoints: usize,
+    /// An ordered map from checkpoint identifier to checkpoint.
+    checkpoints: BTreeMap<C, Checkpoint>,
+    // /// TODO: Add a tree that is used to cache the known roots of subtrees in the "cap" of nodes between
+    // /// `SHARD_HEIGHT` and `DEPTH` that are otherwise not directly represented in the tree.  This
+    // /// cache will be automatically updated when computing roots and witnesses. Leaf nodes are empty
+    // /// because the annotation slot is consistently used to store the subtree hashes at each node.
+    // cap_cache: Tree<Option<Rc<H>>, ()>
```
ebfull, line 1475: Generally speaking I'm not in favor of merging commented code that is not built / tested, but I won't oppose it in this case because it seems like we're on the path to implementing this next anyway.

ebfull (approved):
ACK. I've done at least four passes and haven't found anything majorly concerning and the testing appears very thorough.

PR #52, opened 2023-01-04
Compute witnesses as of checkpoint depth rather than as of a root hash.

nuttycom:
This builds atop #49


bridgetree/src/lib.rs
```diff
@@ -764,24 +758,34 @@ pub enum BridgeTreeError {
...
             saved: BTreeMap::new(),
-            checkpoints: vec![],
+            checkpoints: VecDeque::from(vec![Checkpoint::at_length(0, initial_checkpoint_id)]),
             max_checkpoints,
         }
     }
 
-    /// Removes the oldest checkpoint. Returns true if successful and false if
-    /// there are no checkpoints.
+    /// Removes the oldest checkpoin if there are more than `max_checkpoints`. Returns true if
```
nuttycom, line 775:
```suggestion
    /// Removes the oldest checkpoint if there are more than `max_checkpoints`. Returns true if
```

bridgetree/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+
+- `BridgeTree::witness` now takes a checkpoint depth rather than a root hash to
```
softminus, line 12: @nuttycom says this is so we restrict the domain of the witness-generation function to only work on end-of-block tree states (which correspond to the checkpoints), rather than any arbitrary root hash (which could refer to the state of the tree in the middle of a block).

softminus: Also it makes this consistent between `BridgeTree` (what this is modifying) and `CompleteTree` (the reference implementation for tests).

bridgetree/src/lib.rs
```diff
@@ -616,43 +616,35 @@ impl<'a, H: Hashable + Ord + Clone + 'a> MerkleBridge<H> {
 pub struct Checkpoint {
     /// The number of bridges that will be retained in a rewind.
     bridges_len: usize,
-    /// A flag indicating whether or not the current state of the tree
-    /// had been marked at the time the checkpoint was created.
-    is_marked: bool,
```
softminus, line 621: No longer necessary to store.

bridgetree/src/lib.rs
```diff
@@ -616,43 +616,35 @@ impl<'a, H: Hashable + Ord + Clone + 'a> MerkleBridge<H> {
...
-    /// When a mark is forgotten, if the index of the forgotten mark is <= bridge_idx we
-    /// record it in the current checkpoint so that on rollback, we restore the forgotten
-    /// marks to the BridgeTree's "saved" list. If the mark was newly created since the
-    /// checkpoint, we don't need to remember when we forget it because both the mark
-    /// creation and removal will be reverted in the rollback.
-    forgotten: BTreeMap<Position, usize>,
+    /// When a mark is forgotten, we add it to the checkpoint's forgotten set but
+    /// don't immediately remove it from the `saved` map; that removal occurs when
+    /// the checkpoint is eventually dropped.
+    forgotten: BTreeSet<Position>,
```
softminus, line 627: The set of the positions of notes that were spent within the span of this checkpoint (after the checkpoint was made but before the next checkpoint)

softminus: When we roll back to a checkpoint, we use the forgotten set to restore the set of previously (but will be rolled back!) spent nodes to be part of the set of marked leaves.

bridgetree/src/lib.rs
```diff
@@ -714,9 +697,6 @@ impl Checkpoint {
     // using the specified rewrite function. Used during garbage collection.
     fn rewrite_indices<F: Fn(usize) -> usize>(&mut self, f: F) {
```
softminus, line 709: We have a vector of bridges [A-to-B, B-to-C, C-to-D,...]. When we coalesce this to something like [A-to-C, C-to-D], we previously had to rewrite the usize values in the `forgotten: BTreeMap<Position, usize>`, but now that this is gone, we no longer need to.

softminus: Here, `f` will subtract the number of bridges that have been removed via garbage collection.

bridgetree/src/lib.rs
```diff
@@ -732,8 +712,11 @@ pub struct BridgeTree<H, const DEPTH: u8> {
     /// A map from positions for which we wish to be able to compute a
     /// witness to index in the bridges vector.
     saved: BTreeMap<Position, usize>,
-    /// A stack of bridge indices to which it's possible to rewind directly.
-    checkpoints: Vec<Checkpoint>,
+    /// A deque of bridge indices to which it's possible to rewind directly.
+    /// This deque must be maintained to have a minimum size of 1 and a maximum
+    /// size of `max_checkpoints` in order to correctly maintain mark & rewind
+    /// semantics.
+    checkpoints: VecDeque<Checkpoint>,
```
softminus, line 730: We remove them from the front (the oldest ones) as they fall out of the reorg boundary, and we remove them from the back (the most recently added ones) when there *is* a reorg.

bridgetree/src/lib.rs
```diff
@@ -764,24 +747,34 @@ pub enum BridgeTreeError {
...
+    /// Panics if `max_checkpoints < 1` because mark/rewind logic depends upon the presence
+    /// of checkpoints to function.
     pub fn new(max_checkpoints: usize) -> Self {
+        assert!(max_checkpoints >= 1);
         Self {
             prior_bridges: vec![],
             current_bridge: None,
             saved: BTreeMap::new(),
-            checkpoints: vec![],
+            checkpoints: VecDeque::from(vec![Checkpoint::at_length(0)]),
```
softminus: Always create a checkpoint when the tree is empty

bridgetree/src/lib.rs
```diff
@@ -764,24 +747,34 @@ pub enum BridgeTreeError {
 
 impl<H, const DEPTH: u8> BridgeTree<H, DEPTH> {
     /// Construct an empty BridgeTree value with the specified maximum number of checkpoints.
+    ///
+    /// Panics if `max_checkpoints < 1` because mark/rewind logic depends upon the presence
+    /// of checkpoints to function.
     pub fn new(max_checkpoints: usize) -> Self {
+        assert!(max_checkpoints >= 1);
```
softminus, line 765: Usually the max_checkpoints will be 100, the reorg limit

bridgetree/src/lib.rs
```diff
@@ -764,24 +747,34 @@ pub enum BridgeTreeError {
...
-        if self.checkpoints.is_empty() {
-            false
-        } else {
-            self.checkpoints.remove(0);
+        if self.checkpoints.len() > self.max_checkpoints {
+            let c = self
+                .checkpoints
+                .pop_front()
+                .expect("Checkpoints deque is known to be non-empty.");
+            for pos in c.forgotten.iter() {
```
softminus, line 783: When the checkpoint drops off, we actually remove (in our current tree) the marked nodes that were on the `forgotten` list for that checkpoint.

bridgetree/src/lib.rs
```diff
@@ -969,10 +962,10 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
         self.current_bridge.as_ref().map(|b| b.current_leaf())
     }
 
-    /// Marks the current leaf as one for which we're interested in producing
-    /// a witness. Returns an optional value containing the
-    /// current position if successful or if the current value was already
-    /// marked, or None if the tree is empty.
+    /// Marks the current leaf as one for which we're interested in producing a witness.
```
softminus, line 976: Current leaf = most recently added leaf

bridgetree/src/lib.rs
```diff
@@ -1029,15 +1021,12 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
-                if !c.marked.contains(&position) {
-                    c.forgotten.insert(position, idx);
-                }
-            }
+        if self.saved.contains_key(&position) {
+            let c = self
+                .checkpoints
+                .back_mut()
+                .expect("Checkpoints deque must never be empty.");
+            c.forgotten.insert(position);
```
softminus, line 1040: Rather than removing it from the actual map and adding it to the `forgotten` map, it _only_ adds it to the `forgotten` map (actual removal happens when the checkpoint is removed).

softminus: (not on garbage collection: garbage collection = traverse entire tree and fuse information we don't need anymore. checkpoint removal doesn't trigger a GC)

bridgetree/src/lib.rs
```diff
@@ -1065,10 +1052,10 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
                 }
 
                 self.checkpoints
-                    .push(Checkpoint::at_length(self.prior_bridges.len(), is_marked));
```
softminus, line 1068: this function's changes are just a carryover of the types change

bridgetree/src/lib.rs
```diff
@@ -1083,20 +1070,25 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+            for pos in c.marked {
+                self.saved.remove(&pos);
             }
-            None => false,
+
+            self.prior_bridges.truncate(c.bridges_len);
+            self.current_bridge = self
+                .prior_bridges
+                .last()
+                .map(|b| b.successor(self.saved.contains_key(&b.position())));
```
softminus, line 1107: Create a new successor bridge that starts on the previous one, using the `saved` map (_after_ removing the positions that were marked during the being-reverted checkpoint period).

`.successor` generates a new bridge, and its argument is true if the current node is one that's going to be tracked.

`.map(|b| b.successor(self.saved.contains_key(&b.position())))` operates over a _single_ item, and there's no `Vec` or anything like that, the type that's being mapped over is a `Maybe`.

`saved` map goes from marked node positions to the bridge that ends at that node's position -- since the beginning of time.

Each checkpoint can contain multiple bridges! So rewinds can discard multiple bridges.

Marking a node moves the current bridge to the frozen/immutable `prior_bridges` and creates a new mutable `current_bridge` starting at that position.

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
     /// Returns `None` if there is no available witness to that
     /// position or if the root does not correspond to a checkpointed
     /// root of the tree.
-    pub fn witness(&self, position: Position, as_of_root: &H) -> Option<Vec<H>> {
-        self.witness_inner(position, as_of_root).ok()
-    }
-
-    fn witness_inner(&self, position: Position, as_of_root: &H) -> Result<Vec<H>, WitnessingError> {
+    pub fn witness(
+        &self,
```
softminus, line 1120: self=tree

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+        position: Position,
+        checkpoint_depth: usize,
```
softminus, line 1122: as of `checkpoint_depth` checkpoints back, give us the witness for the node at `position`

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+    ) -> Result<Vec<H>, WitnessingError> {
         #[derive(Debug)]
         enum AuthBase<'a> {
             Current,
             Checkpoint(usize, &'a Checkpoint),
-            NotFound,
```
softminus, line 1117: This PR removes the old functionality where you could pass in an arbitrary tree hash when asking the tree for a witness, so the `NotFound` case is no longer relevant -- `witness` is fed a checkpoint index.

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
-            .filter(|(_, c)| &c.root(&self.prior_bridges, max_alt) == as_of_root)
-            .last()
-            .map(|(i, c)| AuthBase::Checkpoint(i, c))
-            .unwrap_or_else(|| {
-                if self.root(0).as_ref() == Some(as_of_root) {
-                    AuthBase::Current
-                } else {
-                    AuthBase::NotFound
-                }
-            });
```
softminus, line 1139: Removed stuff tries to find a checkpoint with a matching tree hash.

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
+        let auth_base = if checkpoint_depth == 0 {
```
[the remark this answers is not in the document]
softminus: This will _not_ catch the case where the position we're told to generate a witness at has _never_ been marked!

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1097,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
...
-                        // in the same checkpoint
-                        c.forgotten
-                            .get(&position)
-                            .filter(|_| !c.marked.contains(&position))
-                    })
-                } else {
-                    None
-                }
-            })
             .ok_or(WitnessingError::PositionNotMarked(position))?;
```
softminus, line 1160: `saved` map goes from marked node positions to the bridge that ends at that node's position -- since the beginning of time; so if it's not in that map, it's not saved and so we give `WitnessingError::PositionNotMarked(position)`

softminus: In a future commit, we can have the `saved` map lookup happen before the iteration over checkpoints.

softminus: So now we know it didn't get saved _after_ the checkpoint, and we know it got saved, so it has got to have been saved _before_ the checkpoint.

bridgetree/src/lib.rs
```diff
@@ -1198,10 +1179,6 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
                     fuse_from - checkpoint.bridges_len,
                 ))
             }
-            AuthBase::NotFound => {
```
softminus, line 1201: `AuthBase` doesn't have a `NotFound` variant

bridgetree/src/lib.rs
```diff
@@ -1216,17 +1193,7 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
         // checkpoints; we cannot remove information that we might need to restore in
         // a rewind.
         if self.checkpoints.len() == self.max_checkpoints {
-            let gc_len = self.checkpoints.first().unwrap().bridges_len;
-            // Get a list of the leaf positions that we need to retain. This consists of
```
softminus, line 1220: We no longer need to regenerate this list of positions since it's directly out of `saved` with the new semantics.

bridgetree/src/lib.rs
```diff
@@ -1346,8 +1316,8 @@ mod tests {
             BridgeTree::root(self, checkpoint_depth)
         }
 
-        fn witness(&self, position: Position, as_of_root: &H) -> Option<Vec<H>> {
```
softminus, line 1349: Type changes

bridgetree/src/lib.rs
```diff
@@ -1467,6 +1437,22 @@ mod tests {
...
+    fn check_garbage_collect<H: Hashable + Clone + Ord, const DEPTH: u8>(
+        mut tree: BridgeTree<H, DEPTH>,
+    ) {
+        // Add checkpoints until we're sure everything that can be gc'ed will be gc'ed
+        for _ in 0..tree.max_checkpoints {
+            tree.checkpoint();
+        }
+
+        let mut tree_mut = tree.clone();
+        tree_mut.garbage_collect();
```
softminus, line 1469: Check that garbage collection doesn't affect generated witnesses.

bridgetree/src/lib.rs
```diff
@@ -1506,40 +1492,10 @@ mod tests {
...
-    fn drop_oldest_checkpoint() {
-        let mut t = BridgeTree::<String, 6>::new(100);
-        t.checkpoint();
-        t.append("a".to_string());
-        t.mark();
-        t.append("b".to_string());
-        t.append("c".to_string());
-        assert!(
-            t.drop_oldest_checkpoint(),
-            "Checkpoint drop is expected to succeed"
```
softminus, line 1538: This test should be adapted to the new semantics for dropping the oldest checkpoint.

bridgetree/src/lib.rs
```diff
@@ -1562,6 +1518,17 @@ mod tests {
 
     #[test]
     fn garbage_collect() {
+        let mut tree: BridgeTree<String, 7> = BridgeTree::new(100);
+        let empty_root = tree.root(0);
+        tree.append("a".to_string());
+        for _ in 0..100 {
+            tree.checkpoint();
+        }
+        tree.garbage_collect();
```
softminus, line 1549: We test to make sure the garbage collection doesn't prevent us from rewinding

bridgetree/src/lib.rs
```diff
@@ -1588,7 +1555,7 @@ mod tests {
         assert_eq!(t.prior_bridges().len(), 20 + 14 - 2);
         let witnesss = has_witness
             .iter()
-            .map(|pos| match t.witness_inner(*pos, &t.root(0).unwrap()) {
+            .map(|pos| match t.witness(*pos, 0) {
```
softminus, line 1580: New `witness` argument types

bridgetree/src/lib.rs
```diff
@@ -1598,46 +1565,29 @@ mod tests {
...
-                t.witness(*pos, &t.root(0).unwrap())
-                    .expect("Must be able to get auth path")
-            })
+            .map(|pos| t.witness(*pos, 0).expect("Must be able to get auth path"))
             .collect::<Vec<_>>();
         assert_eq!(witnesss, retained_witnesss);
     }
 
-    #[test]
-    fn garbage_collect_idx() {
```
softminus, line 1610: This got put in the start of the `garbage_collect` test

bridgetree/src/lib.rs
```diff
@@ -1662,7 +1612,7 @@ mod tests {
             )
         ) {
             let tree = new_combined_tree(100);
-            check_operations(tree, 4, &ops)?;
```
softminus, line 1665: `DEPTH` method lets it figure out its own depth at runtime

incrementalmerkletree/src/testing.rs
```diff
@@ -64,7 +67,7 @@ pub trait Tree<H> {
     /// Returns `None` if there is no available witness to that
     /// position or if the root does not correspond to a checkpointed
```
softminus, line 59: This is out of date -- witness doesn't take a root anymore!

incrementalmerkletree/src/testing/complete_tree.rs
```diff
@@ -32,214 +36,209 @@ pub(crate) fn root<H: Hashable + Clone>(mut leaves: Vec<H>) -> H {
...
+        }
+    }
+}
+
 #[derive(Clone, Debug)]
-pub struct TreeState<H: Hashable> {
-    leaves: Vec<H>,
-    current_offset: usize,
+pub struct CompleteTree<H, const DEPTH: u8> {
+    leaves: Vec<Option<H>>,
```
softminus, line 64: Leaves being `Option<H>` will enable `shardtree` out of order testing -- not related to empty leaves

softminus: `None` represents lack of information about the leaf in question at that position

incrementalmerkletree/src/testing.rs
softminus: Keep this comment since it's actual code review!!

incrementalmerkletree/src/testing.rs
```diff
@@ -64,17 +58,20 @@ pub trait Tree<H> {
     /// Returns `None` if there is no available witness to that
```
softminus, line 58:
```
/// Returns `None` if there is no available witness to that
/// position or if the root does not correspond to a checkpointed
/// root of the tree.
```
This comment is out of date, since `witness` doesn't take a root hash but instead a checkpoint depth index.

bridgetree/src/lib.rs
```diff
@@ -1506,70 +1512,53 @@ mod tests {
...
-                    tree.witness(*pos, &tree.root(0).unwrap()),
-                    tree_mut.witness(*pos, &tree.root(0).unwrap())
-                );
-            }
+            check_garbage_collect(tree);
         }
     }
 
-    #[test]
-    fn drop_oldest_checkpoint() {
```
softminus, line 1529: I think this test can be adapted to the new semantics for dropping the oldest checkpoint.

softminus (changes requested):
Untested ack, besides the two comments.

bridgetree/src/lib.rs
```diff
@@ -1105,58 +1116,47 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
     /// Returns `None` if there is no available witness to that
```
softminus, line 1116: Likewise, this comment should be changed to not refer to a `root` since the argument type has changed.

softminus (changes requested):
saw a second occurrence of the `witness` comment needing to be changed, and I don't think I can append it to the review, so here goes a second review

bridgetree/src/lib.rs
nuttycom: I looked at this and it cannot, because it's no longer possible to produce a tree state where `drop_oldest_checkpoint` can succeed, because in order to create such a state you have to call `checkpoint`, which calls `drop_oldest_checkpoint` when `max_checkpoints` is exceeded. So the `drop_oldest_checkpoint` function is now just an internal implementation detail of the `checkpoint` function, and can't be tested independently.

softminus (approved):
utACK (previous review comments were addressed with the added commit).

bridgetree/src/lib.rs
softminus: That makes sense; thanks for looking into it!

ebfull (approved):

PR #55, opened 2023-02-16
Update MSRV to 1.60

nuttycom:
The transitive dependency we have on `unarray` is forcing this update. Also, update CI actions to eliminate dependencies on the unmaintained `actions-rs` Rust CI tools.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Run tests
-        run: cargo test --all-features --verbose --release --all
+        run: cargo test --all-features --verbose --workspace
       - name: Verify working directory is clean
         run: git diff --exit-code
 
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v3
       # Build benchmarks to prevent bitrot
       - name: Build benchmarks
-        run: cargo build --all --benches
+        run: cargo build --workspace --benches
 
   doc-links:
     name: Intra-doc links
@@ -33,7 +33,7 @@ jobs:
       - run: cargo fetch
       # Requires #![deny(rustdoc::broken_intra_doc_links)] in crates.
       - name: Check intra-doc links
-        run: cargo doc --all --document-private-items
+        run: cargo doc --workspace --document-private-items
 
   fmt:
     name: Rustfmt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,71 +9,37 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -9,71 +9,37 @@ jobs:
...
-        with:
-          toolchain: 1.56.1
-          override: true
+      - uses: actions/checkout@v3
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --verbose
+        run: cargo test --all-features --verbose --release --all
```
str4d, line 15: - We need `--release` in `zcash/librustzcash` because tests involving ZKPs are too slow in release mode. Here we don't have that restriction, so we should run tests in debug mode for both improved stack traces in the event of a crash, and faster CI.
- `--all` is a deprecated alias for `--workspace` (which we _do_ need here now that this repo is a workspace).
```suggestion
        run: cargo test --all-features --verbose --workspace
```

.github/workflows/ci.yml
```diff
@@ -9,71 +9,37 @@ jobs:
...
-          toolchain: 1.56.1
-          override: true
+      - uses: actions/checkout@v3
       # Build benchmarks to prevent bitrot
       - name: Build benchmarks
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --benches
+        run: cargo build --all --benches
```
str4d, line 26:
```suggestion
        run: cargo build --workspace --benches
```

.github/workflows/ci.yml
```diff
@@ -9,71 +9,37 @@ jobs:
...
-      # Requires #![deny(intra_doc_link_resolution_failure)] in crates.
+      - uses: actions/checkout@v3
+      - run: cargo fetch
+      # Requires #![deny(rustdoc::broken_intra_doc_links)] in crates.
       - name: Check intra-doc links
-        uses: actions-rs/cargo@v1
-        with:
-          command: doc
-          args: --document-private-items
+        run: cargo doc --all --document-private-items
```
str4d, line 36:
```suggestion
        run: cargo doc --workspace --document-private-items
```

.github/workflows/ci.yml
```diff
@@ -9,71 +9,37 @@ jobs:
...
-          toolchain: 1.56.1
-          override: true
-      - run: rustup component add rustfmt
-      - uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: -- --check
+      - uses: actions/checkout@v3
+      - name: Check formatting
+        run: cargo fmt --all -- --check
```
str4d, line 45: Here `--all` is correct.

str4d (changes requested):

.github/workflows/ci.yml
nuttycom: Using `--release` here also makes tests *much* faster; I routinely do this locally because it cuts the time of the proptests from minutes down to seconds.

str4d (approved):
utACK

PR #58, opened 2023-02-23
Fix a bug on 32-bit platforms that caused `Position::root_level` to return the wrong value

### LINKED ISSUE #56: On 32-bit platforms, `Position::root_level` returns the wrong value
In version 0.3.0 this was `Position::max_altitude`.

daira:
fixes #56


nuttycom (approved):
utACK

PR #59, opened 2023-02-24
Do not eagerly track marked nodes in the current bridge.

nuttycom:
This fixes a minor compatibility issue with respect to possible deserialization of the `incrementalmerkletree-0.3.0` `MerkleBridge` type. It is not necessary to track a marked leaf in the current bridge *unless* the marked position is the bridge's starting position and the prior bridge ended at a checkpoint.


bridgetree/src/lib.rs
```diff
@@ -988,8 +987,10 @@ impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
                 {
```
daira, line 987: ^ This could be `|prior_b| prior_b.position() == pos`.

daira (approved):
utACK with non-blocking suggestion.

PR #63, opened 2023-03-22
Make `bridgetree` polymorphic in checkpoint identifier type.

nuttycom:


bridgetree/src/lib.rs
```diff
@@ -655,8 +655,8 @@ impl Checkpoint {
 
     /// The unique identifier for the checkpoint, which is simply an automatically incrementing
     /// index over all checkpoints that have ever been created in the history of the tree.
```
daira, line 657: Update this comment please.

bridgetree/src/lib.rs
```diff
@@ -823,7 +823,7 @@ impl<H, const DEPTH: u8> BridgeTree<H, DEPTH> {
     }
 }
 
-impl<H: Hashable + Ord + Clone, const DEPTH: u8> BridgeTree<H, DEPTH> {
+impl<H: Hashable + Ord + Clone, C: Clone + Ord, const DEPTH: u8> BridgeTree<H, C, DEPTH> {
```
daira, line 826:
```suggestion
impl<H: Hashable + Ord + Clone, C: Ord + Clone, const DEPTH: u8> BridgeTree<H, C, DEPTH> {
```

bridgetree/src/lib.rs
```diff
@@ -714,7 +714,7 @@ impl Checkpoint {
 /// A sparse representation of a Merkle tree with linear appending of leaves that contains enough
 /// information to produce a witness for any `mark`ed leaf.
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct BridgeTree<H, const DEPTH: u8> {
+pub struct BridgeTree<H, C, const DEPTH: u8> {
```
daira, line 717: Rust style question: since much of the functionality of `BridgeTree` is dependent on `H: Hashable + Ord + Clone, C: Clone + Ord`, should that be declared here, rather than only in the impl block that directly depends on it? (If it were only a small number of methods dependent on these constraints, I'd understand; for example nothing depends on `H: Debug, C: Debug` except `fmt`. But for these constraints it appears that `BridgeTree` isn't really usable without them.)

daira (approved):
utACK with comments.

bridgetree/src/lib.rs
nuttycom: The constraints are only relevant to a subset of the operations, so according to the principle of least authority they should not be more broadly scoped. I've found in general that minimally constraining type parameters in struct declarations ends up eliminating awkward spurious constraints elsewhere.

daira (commented):
Post-hoc ACK f7931ec31f7a1ed1ab13c519f9b90d0291bbdfdf

PR #64, opened 2023-04-10
Use `MerklePath` when returning witnesses.

nuttycom:


ebfull (approved):

incrementalmerkletree/src/lib.rs
```diff
@@ -464,6 +464,53 @@ impl<'a> From<&'a Address> for Option<Position> {
...
+
+impl<H: Hashable, const DEPTH: u8> MerklePath<H, DEPTH> {
+    /// Returns the root of the tree corresponding to this path applied to `leaf`.
+    pub fn root(&self, leaf: H) -> H {
+        self.path_elems
+            .iter()
+            .enumerate()
+            .fold(leaf, |root, (i, h)| {
+                let level = Level(i as u8);
+                if self.position.0 >> i & 0x1 == 0 {
```
daira, line 505:
```suggestion
                if (self.position.0 >> i) & 0x1 == 0 {
```
It's correct as-is, but this is clearer I think. Non-blocking.

incrementalmerkletree/src/lib.rs
```diff
@@ -464,6 +464,53 @@ impl<'a> From<&'a Address> for Option<Position> {
...
+}
+
+impl<H: Hashable, const DEPTH: u8> MerklePath<H, DEPTH> {
+    /// Returns the root of the tree corresponding to this path applied to `leaf`.
+    pub fn root(&self, leaf: H) -> H {
+        self.path_elems
+            .iter()
+            .enumerate()
+            .fold(leaf, |root, (i, h)| {
+                let level = Level(i as u8);
```
daira, line 504:
```suggestion
                let level = Level(u8::try_from(i).unwrap());
```

incrementalmerkletree/src/witness.rs
```diff
@@ -251,13 +203,13 @@ impl<H: Hashable + Clone, const DEPTH: u8> IncrementalWitness<H, DEPTH> {
             .enumerate()
         {
             auth_path.push(match p {
-                Some(node) => (node.clone(), true),
-                None => (filler.next(Level::from((i + 1) as u8)), false),
+                Some(node) => node.clone(),
+                None => filler.next(Level::from((i + 1) as u8)),
```
daira, line 207:
```suggestion
                None => filler.next(Level::from(u8::try_from(i + 1).unwrap())),
```

daira (approved):
utACK with non-blocking suggestions.

incrementalmerkletree/src/lib.rs
nuttycom: Does this compile to a cast, or does it have to create an intermediate `Result`? I ask because this is known by construction be safe; `path_elems` is checked to have length equal to `DEPTH` which cannot exceed a U8.

PR #69, opened 2023-05-16
Add support for tracking legacy frontiers and witnesses in `ShardTree` instances.

nuttycom:


dconnolly:
A linter blurp

image

shardtree/src/lib.rs
```diff
@@ -2598,6 +2601,181 @@ where
...
+                                    r_addr
+                                },
+                                truncate_at,
+                            )?)
+                        };
+
+                        // Compute the root value based on the child roots; these may contain the
+                        // hashes of empty/truncated nodes.
+                        let (root, new_left, new_right) = match (l_result, r_result) {
+                            (Some((l_root, new_left)), Some((r_root, new_right))) => (
```
nathan-at-least, line 2903: How is this case possible? Wouldn't it require `r_addr.contains(&target_addr) && r_addr.contains(&target_addr)` which should be impossible?

In fact, before I got to this snippet, I assumed that conjunction was impossible and wondered if we wanted an `Either` instead of a pair of `Option`…

shardtree/src/lib.rs
```diff
@@ -2598,6 +2601,181 @@ where
...
+                            (Some((l_root, new_left)), None) => (l_root, new_left, None),
+                            (None, Some((r_root, new_right))) => (r_root, None, new_right),
+                            (None, None) => unreachable!(),
+                        };
+
+                        let new_parent = Tree(Node::Parent {
+                            ann: new_left
+                                .as_ref()
+                                .and_then(|l| l.node_value())
+                                .zip(new_right.as_ref().and_then(|r| r.node_value()))
```
nathan-at-least, line 2917: `Option::zip` drops either value unless both are present. Can only one of `new_left` or `new_right` be `Some` and in that case is `ann: None` correct?

shardtree/src/lib.rs
```diff
@@ -2772,6 +2958,41 @@ where
         }
     }
 
+    pub fn witness_caching(
```
nathan-at-least, line 3201: API docs for `pub` fn would be helpful for newcomers like me.

shardtree/src/lib.rs
```diff
@@ -711,6 +712,12 @@ impl fmt::Display for InsertionError {
     }
 }
 
+impl std::error::Error for InsertionError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
```
nathan-at-least, line 724: I had thought all `std::error::Error` methods had default impls so this could simply be:

```
impl std::error::Error for InsertionError {}
```

shardtree/src/lib.rs
```diff
@@ -2088,15 +2106,64 @@ pub struct ShardTree<S: ShardStore, const DEPTH: u8, const SHARD_HEIGHT: u8> {
...
+    fn from(err: QueryError) -> Self {
+        ShardTreeError::Query(err)
+    }
+}
+
+impl<S> From<InsertionError> for ShardTreeError<S> {
+    fn from(err: InsertionError) -> Self {
+        ShardTreeError::Insert(err)
+    }
+}
```
nathan-at-least, line 2155: In my personal code, I make heavy use of "boiler-plate removal" dependencies. In this case `derive_more` crate enables `#[derive(derive_more::From)]` on enums that remove this boilerplate. Do we have a policy for or against such crates?

shardtree/src/lib.rs
```diff
@@ -2088,15 +2106,64 @@ pub struct ShardTree<S: ShardStore, const DEPTH: u8, const SHARD_HEIGHT: u8> {
...
+
+impl<S: fmt::Display> fmt::Display for ShardTreeError<S> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self {
+            ShardTreeError::Query(q) => {
+                write!(f, "{}", q)
+            }
+            ShardTreeError::Insert(i) => {
+                write!(f, "{}", i)
+            }
```
nathan-at-least, line 2165: The above two clauses _could_ be written as:

```
  ShardTreeError::Query(q) => q.fmt(f),
  ShardTreeError::Insert(i) => i.fmt(f),
```

I'm not sure how we consense on style stuff. I prefer my suggested rewrite because it is the most direct delegation, and it's also concise code.

shardtree/src/lib.rs
```diff
@@ -725,6 +732,35 @@ pub enum QueryError {
...
+        }
+    }
+}
+
+impl std::error::Error for QueryError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
```
nathan-at-least, line 768: Repeated comment on dropping default method impl.

shardtree/src/lib.rs
nathan-at-least: I just checked that `derive_more::From` can be instructed to ignore the `::Storage` variant (which would otherwise attempt an incoherent blanket impl).

nathan-at-least (approved):
_Partial_ utACK from commits 3d5810fd52e2bf1711eb50f3c126e439c193b936 through 3d5810fd52e2bf1711eb50f3c126e439c193b936.

I'm posting now because I'm anxious about losing state or getting pulled off on other stuff on Monday, but I will attempt to review the remaining commits on Monday.

No blockers. Lots of clarifying questions, some api doc & code style suggestions.

shardtree/src/lib.rs
```diff
@@ -2675,69 +2699,97 @@ impl<
...
+    pub fn truncate_removing_checkpoint(
+        &mut self,
+        checkpoint_id: &C,
+    ) -> Result<bool, ShardTreeError<S::Error>> {
+        match self
+            .store
+            .get_checkpoint(checkpoint_id)
+            .map_err(ShardTreeError::Storage)?
+        {
+            Some(c) => {
```
daira, line 2734: This uses a `match` in the same situation where `truncate_to_depth` uses an `if let Some(...)`. I suggest making them consistent.

shardtree/src/lib.rs
```diff
@@ -2488,7 +2512,7 @@ impl<
         Ok(max_insert_position.map(|p| (p, all_incomplete)))
     }
 
-    /// Insert a tree by decomposing it into its [`SHARD_HEIGHT`] or smaller parts (if necessary)
+    /// Insert a tree by decomposing it into its `SHARD_HEIGHT or smaller parts (if necessary)
```
daira, line 2515:
```suggestion
    /// Insert a tree by decomposing it into its `SHARD_HEIGHT` or smaller parts (if necessary)
```

daira (approved):
utACK 4732ea69edae3b1fc9ad2bf034e2d3040764f9c1 (modulo conflicts).

shardtree/src/lib.rs
nathan-at-least: I believe `clippy` also suggests/errors when a match can be converted to `if let <pat> = …`, and I advocate following all `clippy` and `fmt` (with explicit opt-outs when necessary).

nathan-at-least (approved):
utACK - Reviewed the remaining commits that I didn't cover last week.

nathan-at-least:
ACK w/ local `cargo test` to complement my source review.

shardtree/src/lib.rs
```diff
@@ -3,11 +3,12 @@ use core::convert::TryFrom;
...
 use core::ops::{Deref, Range};
 use either::Either;
-use incrementalmerkletree::frontier::NonEmptyFrontier;
 use std::collections::{BTreeMap, BTreeSet};
 use std::rc::Rc;
 
-use incrementalmerkletree::{Address, Hashable, Level, MerklePath, Position, Retention};
+use incrementalmerkletree::{
+    frontier::NonEmptyFrontier, Address, Hashable, Level, MerklePath, Position, Retention,
+};
```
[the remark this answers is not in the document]
daira: The style I prefer is the style we use in this file:
* section unconditionally importing from std and external crates that are "not ours";
* section unconditionally importing from "our" crates (including the current crate);
* each config-dependent group of imports in its own section.

shardtree/src/lib.rs
```diff
@@ -1321,6 +1416,261 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
         }
     }
 
+    // Constructs a pair of trees contains the leaf and ommers of the given frontier.
```
daira, line 1419:
```suggestion
    // Constructs a pair of trees that contain the leaf and ommers of the given frontier.
```

shardtree/src/lib.rs
```diff
@@ -1532,6 +1581,17 @@ pub trait ShardStore {
     /// provided has level `SHARD_HEIGHT - 1`.
     fn truncate(&mut self, from: Address) -> Result<(), Self::Error>;
 
+    /// A tree that is used to cache the known roots of subtrees in the "cap" of nodes between
```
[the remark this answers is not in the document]
daira: It's not an acronym; the analogy is to a mountain cap. But the notion I have of a cap is a subtree that contains all nodes at a given level and their ancestors — which doesn't seem to be what is meant here. So yes please define this term.

shardtree/src/lib.rs
daira: Also, this is an odd method name. Method names are usually verb phrases (e.g. `cache_witnesses`), not gerunds.

shardtree/src/lib.rs
```diff
@@ -676,6 +696,9 @@ impl fmt::Display for InsertionError {
                 write!(f, "Cannot append out-of-order checkpoint identifier.")
             }
             InsertionError::TreeFull => write!(f, "Note commitment tree is full."),
+            InsertionError::InputMalformed(addr) => {
+                write!(f, "Input malformed for insertion at address {:?}", addr)
+            }
```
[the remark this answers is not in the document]
nuttycom: `thiserror` looks really nice, but I don't think we use it in any of the crates we maintain at this point.

shardtree/src/lib.rs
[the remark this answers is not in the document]
nuttycom: @daira's style is more or less what I usually attempt to follow.

shardtree/src/lib.rs
nuttycom: In order to compute the root, we have to compute the root of the left-hand child and the root of the right-hand child, and combine them. The target address may be at the level of one of our ancestors.

shardtree/src/lib.rs
nuttycom: A root can only be annotated with the stable hash of its child nodes; in the case where `new_left` or `new_right` is incomplete, we can't create such an annotation.

shardtree/src/lib.rs
nuttycom: It's basically "compute the witness, caching any intermediate roots in the cap that are produced along the way." I'll add documentation.

shardtree/src/lib.rs
```diff
@@ -3111,7 +3143,7 @@ mod tests {
 
         let root_addr = Address::from_parts(Level::from(4), 1);
         let tree = LocatedPrunableTree::empty(root_addr);
-        let result = tree.insert_frontier_nodes::<()>(frontier.clone(), Retention::Ephemeral);
+        let result = tree.insert_frontier_nodes::<()>(frontier.clone(), &Retention::Ephemeral);
```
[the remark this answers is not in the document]
nuttycom: This isn't a bitflag; it's an `incrementalmerkletree::Retention` value, which may contain an arbitrarily large checkpoint identifier.

shardtree/src/lib.rs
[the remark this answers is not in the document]
nuttycom: It is now documented in the field definition.

nuttycom:
@daira @nathan-at-least here are the [changes since your previous approvals](https://github.com/zcash/incrementalmerkletree/compare/e57c252..f5889dffa5ccb59a66583c72e410e442875e1151); these are substantial enough to require a re-review if you don't mind.

incrementalmerkletree/Cargo.toml
```diff
@@ -21,5 +21,10 @@ proptest = { version = "1.0.0", optional = true }
 proptest = "1.0.0"
 
 [features]
+# The legacy-api feature guards types and functions that were previously 
+# part of the `zcash_primitives` crate. Those types were removed in the
+# `zcash_primitives` 0.12 release and are now maintained here.
 legacy-api = []
+# The test-depenencies feature guards types and functions that are
```
daira, line 28:
```suggestion
# The test-dependencies feature guards types and functions that are 
```

daira (approved):
utACK, but I did not review new logic with any thoroughness.

shardtree/src/lib.rs
```diff
@@ -3533,17 +3536,128 @@ pub mod testing {
...
+            // not considered incomplete here because they appear
+            // at the tip of the tree.
+            Err(ShardTreeError::Query(QueryError::TreeIncomplete(xs))) if xs == vec![
+                Address::from_parts(Level::from(2), 1),
+                Address::from_parts(Level::from(1), 4),
+            ]
+        );
+
+        assert_matches!(tree.truncate_to_depth(1), Ok(true));
+    }
```
daira, line 3635: There are no checks on the state of `tree` after this call.

nuttycom: Good point, I can add some.

daira (approved):
utACK accb8d7d8d50d4f4c77f5392a59cb2aa5fe17c73 (single commit; caveats about thoroughness of my previous review still apply).

incrementalmerkletree/src/witness.rs
```diff
@@ -84,11 +84,34 @@ impl<H, const DEPTH: u8> IncrementalWitness<H, DEPTH> {
...
+            .witness_addrs(Level::from(DEPTH))
+            .filter_map(|(addr, source)| {
+                if source == Source::Future {
+                    Some(addr)
+                } else {
+                    None
+                }
+            })
+            .take(self.filled.len())
+            .fold(0u64, |acc, addr| acc + 2u64.pow(addr.level().into()));
```
AloeareV, line 105: Another one that's correct, after taking me while to grok..it's a very similar kind of tree traversal to `common_ancestor`, I suspect by the time I read another half-dozen functions in this space it'll start to be intuitive to me.

incrementalmerkletree/src/witness.rs
```diff
@@ -212,6 +235,24 @@ impl<H: Hashable + Clone, const DEPTH: u8> IncrementalWitness<H, DEPTH> {
...
+        let mut base_tree = CommitmentTree::<String, 6>::empty();
+        for c in 'a'..'h' {
+            base_tree.append(c.to_string()).unwrap();
+        }
+        let mut witness = IncrementalWitness::from_tree(base_tree);
+        for c in 'h'..'z' {
+            witness.append(c.to_string()).unwrap();
+        }
+
+        assert_eq!(witness.tip_position(), Position::from(24));
```
AloeareV, line 256: This 24 confused me at first, as it's a 0-indexed 'a' through 'y'...darn exclusive ranges.

AloeareV (commented):
I've reviewed all of the changes not inside `shardtree/src/lib.rs`. I'm partway through reading that file, and it's proving extensive enough I'm gonna break it off into its own review.

A note on terminology: The standard lingo in use here is that the root of the tree is the most-ancestral node. Each non-root node has one parent, and each non-leaf node has two children. Intuitively (from the genetic meaning of the word ancestor), the leaves would be ancestors, and two nodes would be hashed together to create a child node. Hashing two nodes together to create the ancestor of those nodes is the opposite what I expect the term 'ancestor' to mean. 

I don't think this is something that could (or should) be changed, as flipping the meaning of already established terms would just cause more confusion. Maybe it could be better clarified somewhere? 

Edit: The terms ancestor/child is going to get more and more baked in...but as the codebase stands, there's only a few instances of the terms. I don't think the terms have crystallized yet, and I think it would be more intuitive to flip them before they're too baked into the vocabulary.

shardtree/src/lib.rs
```diff
@@ -1868,22 +2323,32 @@ where
...
         let (append_result, position, checkpoint_id) =
-            if let Some(subtree) = self.store.last_shard()? {
+            if let Some(subtree) = self.store.last_shard().map_err(ShardTreeError::Storage)? {
                 if subtree.root.is_complete() {
                     let addr = subtree.root_addr;
 
-                    if addr.index() + 1 >= 0x1 << (SHARD_HEIGHT - 1) {
-                        return Err(InsertionError::TreeFull.into());
-                    } else {
+                    if addr.index() < Self::max_subtree_index() {
```
AloeareV, line 2348: Big readability win

AloeareV (commented):
utACK, this looks great. I'm not sure to draw the line between a long review process where my end result is three sentences, but it's that or add another dozen comments saying "This bit took me a long time to grok but seems correct"

shardtree/src/lib.rs
```diff
@@ -444,6 +475,11 @@ impl<A, V> LocatedTree<A, V> {
         &self.root
     }
 
+    /// Consumes this tree and returns its root as an owned value.
+    pub fn take_root(self) -> Tree<A, V> {
```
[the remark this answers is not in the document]
nuttycom: It's useful to avoid a clone; I added it specifically because I'm using it in https://github.com/zcash/librustzcash/pull/861/commits/afed63f4fc6b102bcb42233c1110939fd226c098#diff-802691a03e5797ebc59e5aa3bd80d3d5c0067754ac0a74f7a7a5666420780d7dR809

zancas:
> I've reviewed all of the changes not inside `shardtree/src/lib.rs`. I'm partway through reading that file, and it's proving extensive enough I'm gonna break it off into its own review.
> 
> A note on terminology: The standard lingo in use here is that the root of the tree is the most-ancestral node. Each non-root node has one parent, and each non-leaf node has two children. Intuitively (from the genetic meaning of the word ancestor), the leaves would be ancestors, and two nodes would be hashed together to create a child node. Hashing two nodes together to create the ancestor of those nodes is the opposite what I expect the term 'ancestor' to mean.
> 
> I don't think this is something that could (or should) be changed, as flipping the meaning of already established terms would just cause more confusion. Maybe it could be better clarified somewhere?
> 
> Edit: The terms ancestor/child is going to get more and more baked in...but as the codebase stands, there's only a few instances of the terms. I don't think the terms have crystallized yet, and I think it would be more intuitive to flip them before they're too baked into the vocabulary.

Hi!   It also seems to me that this use of word `ancestor` is inverted relative to the meaning in genetics/genealogy.

I proposed some commits that would refine terminology and replace `ancestor` with descendant.

https://github.com/nuttycom/incrementalmerkletree/pull/2

Since that proposal changes the public API I `grepped` the librustzcash repo to check for the pattern `-i -e"ancestor"` and didn't see it.  Given that, my guess is that the impact on consumers would be minimal.

nuttycom:
> Hi! It also seems to me that this use of word `ancestor` is inverted relative to the meaning in genetics/genealogy.

The reason that `ancestor` nodes are named as they are is that, in generic tree terminology, the nodes closer to the root are named "parents" and nodes closer to the leaves are named "children". Inverting this relationship would run afoul of a ton of computer science precedent; the fact that in our use case parent nodes are derived via a computation over their children is incidental. So I'm strongly opposed to changing this relationship.

incrementalmerkletree/src/frontier.rs
```diff
@@ -198,21 +203,23 @@ impl<H, const DEPTH: u8> Frontier<H, DEPTH> {
 
     /// Constructs a new frontier from its constituent parts.
     ///
-    /// Returns `None` if the new frontier would exceed the maximum
-    /// allowed depth or if the list of ommers provided is not consistent
-    /// with the position of the leaf.
+    /// Returns `None` if the new frontier would exceed the maximum allowed depth or if the list of
```
ebfull, line 206: This returns `Err`, not `None`, in the failure case.

incrementalmerkletree/src/lib.rs
```diff
@@ -387,6 +425,11 @@ impl Address {
         }
     }
 
+    /// Returns whether this address is the right-hand child of its parent
+    pub fn is_left_child(&self) -> bool {
```
ebfull, line 429: This should say left-hand child.

incrementalmerkletree/src/witness.rs
ebfull, line 105: nonblocking nit: `acc + (1u64 << addr.level().into())` might be more consistent with the other arithmetic in these crates.

shardtree/src/lib.rs
```diff
@@ -1321,6 +1413,260 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+                if let Some(right) = filled.next() {
+                    // once we have a right-hand node, add a parent with the current tree
+                    // as the left-hand sibling
+                    subtree = Tree::parent(
+                        None,
+                        subtree,
+                        Tree::leaf((right.clone(), RetentionFlags::EPHEMERAL)),
+                    );
+                } else {
+                    break;
```
ebfull, line 1522: Tests don't appear to be exercising this scenario.

shardtree/src/lib.rs
```diff
@@ -1321,6 +1413,260 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+                // the current address is a left child, so create a parent with
+                // an empty right-hand tree
+                subtree = Tree::parent(None, subtree, Tree::empty());
+            } else if let Some(left) = ommers.next() {
+                // the current address corresponds to a right child, so create a parent that
+                // takes the left sibling to that child from the ommers
+                subtree =
+                    Tree::parent(None, Tree::leaf((left, RetentionFlags::EPHEMERAL)), subtree);
+            } else {
+                break;
```
ebfull, line 1455: Tests don't appear to be exercising this scenario.

ebfull (changes requested):

AloeareV:
> > Hi! It also seems to me that this use of word `ancestor` is inverted relative to the meaning in genetics/genealogy.
> 
> The reason that `ancestor` nodes are named as they are is that, in generic tree terminology, the nodes closer to the root are named "parents" and nodes closer to the leaves are named "children". Inverting this relationship would run afoul of a ton of computer science precedent; the fact that in our use case parent nodes are derived via a computation over their children is incidental. So I'm strongly opposed to changing this relationship.

Yeah, I suspected as much. I find it extremely confusing for merkle trees, but I'll accept that it's a lesser confusion than using terms opposite to everyone else.

shardtree/src/lib.rs
nuttycom: Fixed in https://github.com/zcash/incrementalmerkletree/pull/69/commits/082109deacf8611ee7917732e19b56158bda96d5

shardtree/src/lib.rs
nuttycom: Fixed in https://github.com/zcash/incrementalmerkletree/pull/69/commits/082109deacf8611ee7917732e19b56158bda96d5

incrementalmerkletree/src/witness.rs
nuttycom: Fixed in https://github.com/zcash/incrementalmerkletree/pull/69/commits/701d3e6311bafd2bd21bb8f44c8624201d418832

incrementalmerkletree/src/lib.rs
nuttycom: Fixed in https://github.com/zcash/incrementalmerkletree/pull/69/commits/701d3e6311bafd2bd21bb8f44c8624201d418832

incrementalmerkletree/src/frontier.rs
nuttycom: fixed in https://github.com/zcash/incrementalmerkletree/pull/69/commits/701d3e6311bafd2bd21bb8f44c8624201d418832

ebfull (approved):

PR #72, opened 2023-05-19
Remove initial checkpoints.

nuttycom:
The choice to use an initial checkpoint as an anchor for removals in
`unmark` operations produced an awkward situation where it was necessary
to choose a checkpoint identifier just to construct an empty tree. This
removes that decision in favor of eagerly removing data if there is no
checkpoint to tie the removal to.

Builds upon #71


shardtree/src/lib.rs
```diff
@@ -2031,19 +2027,13 @@ where
...
-                    })
-                    .is_err()
-                {
-                    return Ok(false);
-                }
+            if let Some((replacement, pos)) = go(subtree.root_addr, &subtree.root) {
+                self.store.put_shard(LocatedTree {
+                    root_addr: subtree.root_addr,
+                    root: replacement,
+                })?;
```
str4d, line 2034: Is this change a bugfix?

shardtree/src/lib.rs
```diff
@@ -2361,31 +2347,43 @@ where
...
+                        self.store
+                            .get_shard(Self::subtree_addr(position))?
+                            .map(|subtree| {
+                                subtree.clear_flags(BTreeMap::from([(
+                                    position,
+                                    RetentionFlags::MARKED,
+                                )]))
+                            });
+                    if let Some(cleared) = cleared {
+                        self.store.put_shard(cleared)?;
```
str4d, line 2377: To clarify my understanding, this is:
- Fetching the subtree shard containing the marked position.
- If it exists, returning a clone of the shard that has the marked position cleared.
- If _that_ exists, overwrites the previous subtree shard (the reference to which has just been dropped) with the cleared version (which is guaranteed to have had something cleared, because all this logic is gated on `self.get_marked_leaf(position)?.is_some()`).

Related nit: the documentation for `LocatedPrunableTree::clear_flags` says it clears the given position, but it does not mutate through `&self` and instead returns a clone of the tree with the given position cleared (if necessary).

shardtree/src/lib.rs
```diff
@@ -2361,31 +2347,43 @@ where
...
+                        Ok(true)
+                    } else {
+                        Ok(false)
```
str4d, line 2380: And this failure case occurs if we need to clear the mark directly, but do not have the subtree in which it should be contained? It would help for the method documentation to clarify what the `bool` in the return type means.

str4d (approved):
utACK with some questions and clarification requests.

bridgetree/src/lib.rs
```diff
@@ -470,13 +470,13 @@ impl<H, C, const DEPTH: u8> BridgeTree<H, C, DEPTH> {
     ///
     /// Panics if `max_checkpoints < 1` because mark/rewind logic depends upon the presence
     /// of checkpoints to function.
```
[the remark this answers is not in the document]
nuttycom: If `max_checkpoints` is set to 0, it would essentially disable the checkpoint and rewind functionality. I suspect that there are a number of assumptions that it's possible to create a checkpoint scattered throughout the code, so I think that the restriction should be left in place.

shardtree/src/lib.rs
nuttycom: Yes.

shardtree/src/lib.rs
nuttycom: It would actually be an error if the position were indicated as marked but we were unable to retrieve the shard, so I've updated the code to simplify the number of cases here.

ebfull:
I reviewed `2b8e2d62fa599e4b73c166a1c8c36df7ab78804a`, LGTM. Didn't carefully look at the rest of the PR.

PR #74, opened 2023-05-24
Fix incorrect documentation and `cargo doc` issues.

nuttycom:


shardtree/src/lib.rs
```diff
@@ -11,7 +11,10 @@ use incrementalmerkletree::{Address, Hashable, Level, MerklePath, Position, Rete
 bitflags! {
     pub struct RetentionFlags: u8 {
         /// An leaf with `EPHEMERAL` retention can be pruned as soon as we are certain that it is not part
-        /// of the witness for a leaf with `CHECKPOINT` or `MARKED` retention.
+        /// of the witness for a leaf with [`CHECKPOINT`] or `MARKED` retention.
```
daira, line 14:
```suggestion
        /// of the witness for a leaf with [`CHECKPOINT`] or [`MARKED`] retention.
```

daira (approved):
ACK with suggestion. Doc-only change.

PR #75, opened 2023-06-05
Release of `bridgetree` v0.2.1

nuttycom:
The `bridgetree-v0.2.0` tag was created with respect to an invalid repository state, which did not actually contain the `bridgetree` sources after they were factored out from the `incrementalmerkletree` crate.


nuttycom:
superseded by #77

PR #76, opened 2023-06-05
Remove the `Debug` bound from `Hashable`

nuttycom:
The `Debug` bound here is handy but not required for the operations on the trait. Similarly, the `Sized` bound is only required for the `empty_root` operation.


daira (approved):
utACK

PR #78, opened 2023-06-10
`incrementalmerkletree` API docs for location and node relations.

nathan-at-least:
The crate docs still lack an overview of the other pieces, but this felt like a good way to quickly sort out the different abstractions at a glance.

### DIFF
--- a/incrementalmerkletree/src/lib.rs
+++ b/incrementalmerkletree/src/lib.rs
@@ -33,8 +33,7 @@
 //! - `root` - the single node with the largest [Level].
 //! - `ancestor` - the parent or an ancestor of the parent; the root is an ancestor of every other node.
 //!
-//! Note: when refering to `emmers` (plural) we typically mean the node's emmer, plus each ancestor's emmer.
-
+//! Note: we often refer to `emmers` (plural) when describing leaf-to-root paths, so in that context `emmers` refers to the node's emmer, plus each ancestor's emmer.
 
 use either::Either;
 use std::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


incrementalmerkletree/src/lib.rs
```diff
@@ -1,4 +1,39 @@
...
+//!
+//! **Relationship Abstractions:**
+//!
+//! A given node has these navigational relationships:
+//!
+//! - `parent` - the node directly above at one higher [Level].
+//! - `child` - the complementary relationship to parent; a parent may have up to two children because only binary trees are supported.
+//! - `sibling` - the other node with the same parent.
+//! - `cousin` - a node at the same [Level] excluding the sibling.
+//! - `emmer` - the parent's sibling.
```
nuttycom, line 32:
```suggestion
//! - `ommer` - the parent's sibling.
```

incrementalmerkletree/src/lib.rs
```diff
@@ -1,4 +1,39 @@
...
+//! - `root` - the single node with the largest [Level].
+//! - `ancestor` - the parent or an ancestor of the parent; the root is an ancestor of every other node.
+//!
+//! Note: we often refer to `emmers` (plural) when describing leaf-to-root paths, so in that context `emmers` refers to the node's emmer, plus each ancestor's emmer.
```
nuttycom, line 36:
```suggestion
//! Note: we often refer to `ommers` (plural) when describing leaf-to-root paths, so in that context `ommers` refers to the node's ommer, plus each ancestor's ommer.
```

nuttycom (approved):
ACK, doc-only.

PR #81, opened 2023-07-05
Refactor codebase into submodules

str4d:
The external API structure doesn't change, apart from `MemoryShardStore` moving into a submodule. This is primarily to improve maintainability.


nuttycom (approved):
ACK, move-only.

PR #82, opened 2023-07-06
`shardtree` bugfixes

str4d:


nuttycom (approved):
utACK

daira (approved):
utACK

shardtree/src/prunable.rs
```diff
@@ -967,6 +968,22 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
                     }
                 }
 
+                while cur.root_addr.level() + 1 < root_addr.level() {
```
nuttycom, line 971: nonblocking: the comments on line 944 and 964 are now slightly misleading, because they don't acknowledge this block. Also, since this is identical to the loop on 921 modulo the bounds being checked, it'd be good to factor it out into another internal helper function. If you choose to do so, consider also making that function able to handle the loop on 998; the only difference is whether the `nil`s go on the right or the left.

nuttycom (approved):
utACK with suggestions

shardtree/src/prunable.rs
nuttycom: If you act on this, add it as an additional commit since it's a refactoring that shouldn't be mixed with the bugfix.

shardtree/src/prunable.rs
```diff
@@ -905,6 +905,35 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+        /// `expect_left_child` is set to a constant at each callsite, to ensure that this
+        /// function is only called on either the left-most or right-most subtree.
+        fn combine_with_empty<H: Hashable + Clone + PartialEq>(
+            root: LocatedPrunableTree<H>,
+            expect_left_child: bool,
+            incomplete: &mut Vec<IncompleteAt>,
+            contains_marked: bool,
+            prune_below: Level,
+        ) -> LocatedPrunableTree<H> {
+            let sibling_addr = root.root_addr.sibling();
```
nuttycom, line 920: Maybe panic earlier:
```suggestion
            assert_eq!(expect_left_child, root.root_addr.is_left_child);
            let sibling_addr = root.root_addr.sibling();
```

shardtree/src/prunable.rs
```diff
@@ -905,6 +905,35 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+            });
+            let sibling = LocatedTree {
+                root_addr: sibling_addr,
+                root: Tree(Node::Nil),
+            };
+            let (lroot, rroot) = match (expect_left_child, root.root_addr.is_left_child()) {
+                (true, true) => (root, sibling),
+                (false, false) => (sibling, root),
+                _ => panic!("combine_with_empty called on a middle subtree"),
+            };
```
nuttycom, line 933:
```suggestion
            let (lroot, rroot) = if root.root_addr.is_left_child() {
                (root, sibling)
            } else {
                (sibling, root)
            }
```

nuttycom (approved):
utACK with nonblocking suggestion

nuttycom (approved):
utACK 52b8de3

PR #83, opened 2023-07-06
the Position type is exclusively used with leaves, propose a readabil…

zancas:
This proposes a readability enhancing rename to `LeafPosition`


daira (commented):
I'm -1 on this change. It's a breaking change to a public API, and I don't think there's really any ambiguity in the name `Position`. The Zcash protocol spec just uses "position" or "note position".

str4d:
Also -1. "Note position" only makes sense in the context of Zcash, and "position" is never used to refer to anything but leaves. For interior nodes we use "index" to refer to their horizontal distance from the left side of the tree (equivalent of "position" for leaves), but interior nodes can't be referenced without a full `Address` (whereas leaves have an implicit level of 0).

PR #85, opened 2023-07-14
Add `CachingShardStore` for batching writes to a backend `ShardStore`

str4d:


shardtree/src/caching.rs
```diff
@@ -0,0 +1,1253 @@
...
+        Ok(Self {
+            backend,
+            cache,
+            deferred_actions: vec![],
+        })
+    }
+
+    /// Flushes the current cache state to the backend and returns it.
+    pub fn flush(mut self) -> Result<S, S::Error> {
+        for action in &self.deferred_actions {
```
ebfull, line 67: These deferred actions could involve removing checkpoints in the backend that haven't been added to the backend yet (which happens later in this function). Or am I misunderstanding something?

str4d: The deferred actions are still applied live to `cache`; we just need to remember them so we can also apply them to `backend` (as we can't easily determine what needs removing just from the state of `cache`.

ebfull (approved):

shardtree/Cargo.toml
```diff
@@ -36,4 +36,5 @@ test-dependencies = ["proptest", "assert_matches"]
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.9", features = ["criterion", "flamegraph"] } # MSRV 1.56
+dashmap = ">=5, <5.5.0"
```
daira, line 39: What was the issue with 5.5.0?

str4d: As I documented in the commit message, this was to keep MSRV. `dashmap 5.5.0` bumped dependencies in a way that breaks the MSRV of this crate, but because we only use it as a dev-dependency we can just pin in this way. We've had this problem a lot with the `criterion` / `pprof` stack, and it's the standard way we solve it.

daira (commented):
Post-hoc ACK ~~with a question about the dashmap version constraint~~.

shardtree/Cargo.toml
daira: Ah ok. I knew that zcashd has an MSRV of 1.69, the librustzcash crates' MSRVs are 1.65, and dashmap 5.5.0's MSRV is 1.64. I guess it's https://github.com/xacrimon/dashmap/commit/d36311e6340800d3f2c2bdf0f34f72bd234dfa2b that causes the problem? Is there a reason why we can't/shouldn't update the incrementalmerkletree repo's crates to match librustzcash?

In any case,
* shardtree does not appear to declare its MSRV;
* there isn't a comment in this file saying what the MSRV of dashmap is, unlike for pprof and inferno.

str4d: You're correct that `shardtree` doesn't yet declare its MSRV; we should add that. The rest of the crates in this repo have MSRV 1.60, which is what CI is enforcing.

PR #88, opened 2023-07-25
shardtree: Pass checkpoints alongside tree to `ShardTree::insert_tree`

str4d:
There are no `ShardTree` APIs that allow for checkpointing at arbitrary positions, which makes sense as the relevant leaf may have been pruned. For this API to be a viable substitute for `ShardTree::batch_insert`, it therefore needs to support marking any checkpoints associated with the inserted tree.


ebfull (approved):

daira (commented):
Post-hoc utACK.

PR #90, opened 2023-07-25
`shardtree`: Improve documentation

str4d:


shardtree/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+//! `shardtree` is a space-efficient fixed-depth Merkle tree that supports:
+//! - Out-of-order insertion.
+//! - Witnessing of marked leaves.
+//! - Checkpointing, and state restoration to a checkpoint.
```
str4d, line 4: I'd love to have suggestions for more things to put here, but that doesn't have to block the release.

ebfull (approved):

PR #93, opened 2023-11-02
shardtree: Add `root_at_checkpoint_id` methods.

nuttycom:
It can be useful to refer to a specific checkpoint, rather than just a checkpoint depth, for which one wishes to compute the root.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -641,7 +641,7 @@ impl<
     /// given position, and parents of such nodes, will be replaced by the empty root for the
     /// associated level.
     ///
-    /// Use [`Self::root_at_checkpoint_id`] to obtain the root of the overall tree.
+    /// Use [`Self::root_at_checkpoint_depth`] to obtain the root of the overall tree.
     pub fn root(
         &self,
         address: Address,
@@ -1032,17 +1032,17 @@ impl<
         )
     }
 
-    fn witness_internal(
-        &self,
+    fn witness_helper<Ctx>(
+        mut ctx: Ctx,
         position: Position,
         as_of: Position,
+        get_shard: impl Fn(&Ctx, Address) -> Result<Option<LocatedPrunableTree<H>>, S::Error>,
+        mut root: impl FnMut(&mut Ctx, Address, Position) -> Result<H, ShardTreeError<S::Error>>,
     ) -> Result<MerklePath<H, DEPTH>, ShardTreeError<S::Error>> {
         let subtree_addr = Self::subtree_addr(position);
 
         // compute the witness for the specified position up to the subtree root
-        let mut witness = self
-            .store
-            .get_shard(subtree_addr)
+        let mut witness = get_shard(&ctx, subtree_addr)
             .map_err(ShardTreeError::Storage)?
             .map_or_else(
                 || Err(QueryError::TreeIncomplete(vec![subtree_addr])),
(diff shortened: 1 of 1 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.60.0"
+channel = "1.64.0"
```
str4d, line 2: You've bumped the toolchain version without an explicit MSRV bump for the crates. What is this needed for? If an MSRV bump is necessary, it should be done in a separate initial commit (and have corresponding changelog entries like we do in `zcash/librustzcash`).

shardtree/src/lib.rs
```diff
@@ -641,7 +641,7 @@ impl<
     /// given position, and parents of such nodes, will be replaced by the empty root for the
     /// associated level.
     ///
-    /// Use [`Self::root_at_checkpoint`] to obtain the root of the overall tree.
+    /// Use [`Self::root_at_checkpoint_id`] to obtain the root of the overall tree.
```
str4d, line 644: Why was this changed to use the new `*_id` method, rather than being simply a rename to `*_depth`? The latter seems preferable to use for this purpose, as you don't need to know the checkpoint ID.

shardtree/src/lib.rs
```diff
@@ -995,92 +1032,154 @@ impl<
...
+        // compute the remaining parts of the witness up to the root
+        let root_addr = Self::root_addr();
+        let mut cur_addr = subtree_addr;
+        while cur_addr != root_addr {
+            witness.push(self.root_caching(cur_addr.sibling(), as_of + 1)?);
+            cur_addr = cur_addr.parent();
+        }
+
+        Ok(MerklePath::from_parts(witness, position).unwrap())
+    }
```
str4d, line 1093:
```suggestion
    fn witness_helper<Ctx>(
        mut ctx: Ctx,
        position: Position,
        as_of: Position,
        get_shard: impl Fn(&Ctx, Address) -> Result<Option<LocatedPrunableTree<H>>, S::Error>,
        mut root: impl FnMut(&mut Ctx, Address, Position) -> Result<H, ShardTreeError<S::Error>>,
    ) -> Result<MerklePath<H, DEPTH>, ShardTreeError<S::Error>> {
        let subtree_addr = Self::subtree_addr(position);

        // compute the witness for the specified position up to the subtree root
        let mut witness = get_shard(&ctx, subtree_addr)
            .map_err(ShardTreeError::Storage)?
            .map_or_else(
                || Err(QueryError::TreeIncomplete(vec![subtree_addr])),
                |subtree| subtree.witness(position, as_of + 1),
            )?;

        // compute the remaining parts of the witness up to the root
        let root_addr = Self::root_addr();
        let mut cur_addr = subtree_addr;
        while cur_addr != root_addr {
            witness.push(root(&mut ctx, cur_addr.sibling(), as_of + 1)?);
            cur_addr = cur_addr.parent();
        }

        Ok(MerklePath::from_parts(witness, position).unwrap())
    }

    fn witness_internal(
        &self,
        position: Position,
        as_of: Position,
    ) -> Result<MerklePath<H, DEPTH>, ShardTreeError<S::Error>> {
        Self::witness_helper(
            self,
            position,
            as_of,
            |ctx, shard_root| ctx.store.get_shard(shard_root),
            |ctx, address, truncate_at| ctx.root(address, truncate_at),
        )
    }

    fn witness_internal_caching(
        &mut self,
        position: Position,
        as_of: Position,
    ) -> Result<MerklePath<H, DEPTH>, ShardTreeError<S::Error>> {
        Self::witness_helper(
            self,
            position,
            as_of,
            |ctx, shard_root| ctx.store.get_shard(shard_root),
            |ctx, address, truncate_at| ctx.root_caching(address, truncate_at),
        )
    }
```

shardtree/src/lib.rs
```diff
@@ -995,92 +1032,154 @@ impl<
...
-            let mut cur_addr = subtree_addr;
-            while cur_addr != root_addr {
-                witness.push(self.root_caching(cur_addr.sibling(), max_leaf_position + 1)?);
-                cur_addr = cur_addr.parent();
-            }
+    /// Computes the witness for the leaf at the specified position, as of the given checkpoint.
+    ///
+    /// Returns the witness as of the most recently appended leaf if `checkpoint_depth == 0`. Note
+    /// that if the most recently appended leaf is also a checkpoint, this will return the same
+    /// result as `checkpoint_depth == 1`.
```
str4d, line 1149:
```suggestion
```
(but maybe we want to say what happens if the given checkpoint doesn't exist).

str4d (commented):
Reviewed as of fbcaec2ab16d33b1c55a6920914150f560ed1669.

shardtree/src/lib.rs
nuttycom: A dependency bumped its MSRV in a point change, so `main` no longer builds.

rust-toolchain.toml
str4d: We should fix this the same way we have in `zcash/librustzcash`: check in a working `Cargo.lock`.

shardtree/CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to Rust's notion of
 
 ## Unreleased
 
+## Added
+* `Shardtree::{root_at_checkpoint_id, root_at_checkpoint_id_caching}`
+* `Shardtree::{witness_at_checkpoint_id, witness_at_checkpoint_id_caching}`
+
+## Changed
+* Updated MSRV to 1.64.0
```
str4d, line 15: Given that this MSRV bump is not necessary for anything in this PR, I now think we should not bump it, and instead check in a `Cargo.lock` that works, as we do in `zcash/librustzcash`.

str4d (approved):
utACK c9a876f32baace0389d577133659e2c474c9b5fd, with a soft request to not bump MSRV given that it is not necessary for the PR's changes.

shardtree/CHANGELOG.md
nuttycom: So, what's the best way to generate such a Cargo.lock, given that the reason I made this bump in the first place was that I was getting dependency errors?

nuttycom: Fixed.

PR #95, opened 2024-01-19
shardtree: Add the ability to avoid pruning specific checkpoints.

nuttycom:


shardtree/src/lib.rs
```diff
@@ -111,6 +114,12 @@ impl<
         (0x1 << (DEPTH - SHARD_HEIGHT)) - 1
     }
 
+    /// Adds the provided checkpoint to the set of checkpoints to be retained
+    /// across pruning operations.
```
str4d, line 118: This API needs a bit more documentation, because it is unclear whether or not this data being stored in `ShardTree` means it needs to be serialized. In particular, we should document that this API needs to be called after constructing `ShardTree` but before calling any methods that take `&mut self` (or whatever other criterion gates potential pruning). An example would be ideal:
```suggestion
    /// Adds the provided checkpoint to the set of checkpoints to be retained
    /// across pruning operations.
    ///
    /// TODO: A bit more documentation.
    ///
    /// # Examples
    ///
    /// ```
    /// use incrementalmerkletree::Retention;
    /// use shardtree::ShardTree;
    ///
    /// # fn load_shardtree_from_disk() -> ShardTree<SomeShardStore> {
    /// #     todo!("Kris returns something useful for testing here.")
    /// # }
    /// # fn checkpoints_to_retain() -> Vec<SomeCheckpointId> {
    /// #     vec![] // TODO: Return something useful for testing here.
    /// # }
    /// # fn get_next_leaf() -> SomeLeafType {
    /// #     todo!("Kris returns something useful for testing here.")
    /// # }
    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
    /// let mut tree = load_shardtree_from_disk();
    ///
    /// // The required checkpoints are stored separately from the tree.
    /// for checkpoint_id in checkpoints_to_retain() {
    ///     tree.ensure_retained(checkpoint_id);
    /// }
    ///
    /// // Now perform operations that might prune the tree.
    /// tree.append(get_next_leaf(), Retention::Ephemeral)?;
    /// # Ok(())
    /// # }
    /// ```
```

shardtree/src/lib.rs
```diff
@@ -442,10 +451,11 @@ impl<
             checkpoint_count,
             self.max_checkpoints,
```
str4d, line 452: Update this trace to mention `self.to_retain.len()`.

shardtree/src/lib.rs
```diff
@@ -442,10 +451,11 @@ impl<
...
         );
-        if checkpoint_count > self.max_checkpoints {
+        let retain_count = self.max_checkpoints + self.to_retain.len();
+        if checkpoint_count > retain_count {
```
str4d, line 455: This doesn't match my intuition. I expected that only checkpoints in `self.to_retain` that fall outside the range of `self.max_checkpoints` would increase the cap. Instead, this excludes all explicitly-retained checkpoints from the cap, leading to (in the context of Zcash) older checkpoint heights being retained than is necessary.

I don't particularly mind doing it this way if it makes the implementation easier, but a) we should explicitly decide which way we want to go (as I disagree with us changing the behaviour of the new API later in a non-compilation-breaking way; going from the current implementation to the one matching my intuition would constitute data loss to someone depending on the current behaviour), and b) we should document it on `ShardTree::ensure_retained`.

shardtree/src/lib.rs
```diff
@@ -454,8 +464,10 @@ impl<
                     // When removing is true, we are iterating through the range of
                     // checkpoints being removed. When remove is false, we are
                     // iterating through the range of checkpoints that are being
-                    // retained.
-                    let removing = checkpoints_to_delete.len() < remove_count;
+                    // retained, or skipping over a particular checkpoint that we
+                    // have been explicitly asked to retain.
+                    let removing =
+                        checkpoints_to_delete.len() < remove_count && !self.to_retain.contains(cid);
```
str4d, line 470: Does the `clear_at` closure below behave correctly after this change? Its current implementation at a glance appears to assume that we only have a single transition from `removing = true` to `removing = false` when iterating in ascending order through the checkpoints (because it treats `removing = false` as a fixup operation, but `removing = true` as unconditional). We can now have arbitrarily many transitions between those two states, up until we reack the `self.max_checkpoints` boundary.

shardtree/src/lib.rs
```diff
@@ -1355,6 +1367,42 @@ mod tests {
...
+        // to be pruned (but will not, because we explicitly retained it.)
+        for i in 14..24 {
+            assert_eq!(tree.checkpoint(i), Ok(true));
+        }
+
+        // Verify that we can still compute the same root
+        assert_matches!(
+            tree.witness_at_checkpoint_id(Position::from(4), &12),
+            Ok(w) if w == e_witness_12
+        );
```
str4d, line 1405: Add a test case where we request to retain a checkpoint that is within the range of checkpoints that would already be retained. This is to exercise the `self.max_checkpoints + self.to_retain.len()` behaviour I questioned above; we should have a case that checks whichever behaviour we go with.

str4d (changes requested):
Reviewed 34b1a5f704719436f436d0c8275c63e03942bce5.

shardtree/src/lib.rs
nuttycom: The reason to retain all the checkpoints is so that the 100-block rollback limit is matched by the number of checkpoints to which you can rewind, because we use checkpoints not just for witnessing but for rewinds.

shardtree/src/lib.rs
nuttycom: Yes, I believe so. Essentially, the `else` branch only becomes relevant when some prior iteration of the loop has removed data that needs to be restored, and what we're doing in this PR is explicitly skipping over some of those removals.

shardtree/src/lib.rs
nuttycom: I guess that since checkpoint identifiers are intended to be totally ordered, we could partition the set of retained checkpoints and then on each iteration, check whether we're down to the max number of checkpoints, not counting the number of earlier retained checkpoints. That might be okay.

shardtree/src/lib.rs
nuttycom: Hm, I guess maybe I should actually add this to the `ShardStore` interface, so that it's up to the implementer of that to decide whether or not the explicitly retained items are represented in the serialization. In the `zcash_client_sqlite` database, they'll actually be serialized; the retained checkpoint identifier will always be the block height prior to the earliest unscanned `FoundNote` or `ChainTip` range, and each time we go to perform a `put_blocks` operation, we'll check whether the operation has increased that height. 

It gets a little complicated, because `Historic` and `OpenAdjacent` scanning can find notes at _lower_ heights than the current "high water mark" - in the event that that happens, the value of those notes can't be added to the available balance until their entire subtrees are scanned. This leads to an uncomfortable situation where you might be able to increase the available balance if you were to choose a _lower_ checkpoint height, for example if you have a small spendable note at the chain tip, but then discover a large spendable note at a lower height.

shardtree/src/lib.rs
nuttycom: Aha. The better way to do this is, instead of `max_checkpoints`, it should be `min_checkpoints_to_retain`. That would also allow the implementation to skip pruning entirely sometimes in order to batch pruning operations and so forth, and better reflects the semantics we actually want.

shardtree/src/lib.rs
```diff
@@ -66,8 +66,8 @@ mod legacy;
 pub struct ShardTree<S: ShardStore, const DEPTH: u8, const SHARD_HEIGHT: u8> {
     /// The vector of tree shards.
     store: S,
-    /// The maximum number of checkpoints to retain before pruning.
-    max_checkpoints: usize,
+    /// The minumum number of checkpoints to retain when pruning.
```
daira, line 69:
```suggestion
    /// The minimum number of checkpoints to retain when pruning.
```

shardtree/src/lib.rs
```diff
@@ -79,10 +79,10 @@ impl<
     > ShardTree<S, DEPTH, SHARD_HEIGHT>
 {
     /// Creates a new empty tree.
-    pub fn new(store: S, max_checkpoints: usize) -> Self {
+    pub fn new(store: S, min_checkpoints_to_retain: usize) -> Self {
```
daira, line 82: This seems quite a significant semantic change for something that retains the same type signature. Is it the case that previous callers will be correct for the new semantics? In any case please document the parameter.

daira (approved):
utACK with comments.

nuttycom:
Superseded by #184, which reaches the same explicit-checkpoint-retention design rebased onto current `main` and simplified — record retention is decoupled from leaf marking via `ShardTree::ensure_retained`, and no `Reference`/`Marking` changes are required (the retained checkpoint's own `CHECKPOINT` retention protects the witness-supporting nodes). Closing in favor of #184.

— Closed with Claude Code assistance.

PR #103, opened 2024-04-03
shardtree: Ensure `test-dependencies` feature enables `incrementalmerkletree/test-dependencies`

nuttycom:
This is intended to fix an issue with the `docs.rs` build.


str4d (approved):
utACK

PR #106, opened 2024-06-26
shardtree: Use `Tree` constructors instead of struct creation

str4d:
This makes it easier for us to track where new tree nodes are created, and how the node kinds are used within the `ShardTree` data model.


shardtree/src/prunable.rs
```diff
@@ -314,26 +312,20 @@ impl<H: Hashable + Clone + PartialEq> PrunableTree<H> {
...
-                Tree(
-                    Node::Leaf {
-                        value: (H::combine(level, &lv.0, &rv.0), rv.1),
-                    },
-                )
+                Tree::leaf((H::combine(level, &lv.0, &rv.0), rv.1))
             }
-            (left, right) => Tree(
-                Node::Parent {
+            (left, right) => Tree::parent(
```
nuttycom, line 325: I wonder why this doesn't format as a single line?

nuttycom (approved):
utACK 020a7d76d77cfa04c32dd35db1ae109f70155d78

PR #107, opened 2024-06-26
shardtree: Use annotated parent nodes with empty children for inserted frontier ommers.

nuttycom:


shardtree/src/prunable.rs
```diff
@@ -721,21 +711,13 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
-            "Current shard"
-        );
         let LocatedTree { root_addr, root } = self;
         if root_addr.contains(&subtree.root_addr) {
             go(*root_addr, root, subtree, contains_marked).map(|(root, incomplete)| {
                 let new_tree = LocatedTree {
                     root_addr: *root_addr,
                     root,
                 };
-                assert!(new_tree.max_position() >= max_position);
```
nuttycom, line 738: The removal of this assertion is intentional; given the way that `shardtree` was used prior to this PR, it is not possible to determine whether a `Node::Leaf` value in a persisted tree was present as a pruned subtree root, or because of direct insertion as a shard root or the ommer of an inserted frontier. In general, the maximum position of populated nodes within a subtree should only be used when considering the maximum inserted position of the overall tree.

shardtree/src/tree.rs
```diff
@@ -148,6 +148,15 @@ impl<A, V> Tree<A, V> {
         matches!(&self.0, Node::Leaf { .. })
     }
 
+    /// Returns `true` if no additional nodes can be appended to this tree.
+    pub fn is_full(&self) -> bool {
+        match &self.0 {
+            Node::Nil => false,
+            Node::Leaf { .. } | Node::Pruned => true,
+            Node::Parent { right, .. } => right.is_full(),
```
str4d, line 156: Hmm, this definition of `is_full` is a little confusing.
- In the abstract context, "append" is not well-defined, as leaves can be added anywhere we have a `Node::Nil`; tree "fullness" would mean "absence of any `Node::Nil`s". In the concrete context, what we specifically care about is append-on-the-right.
- `Node::Nil` in the abstract context means "empty tree", implying that we can abstractly add nodes below it. But in the concrete context, it means "we have no information", and so it _might_ be the case that no additional nodes can be appended to this tree.

I think it would be clearer if this was named `is_definitely_full_on_right`, and the documentation clarified that returning `false` doesn't mean that we can definitely add more nodes on the right to the tree that this `Tree` represents, only that we have not observed it to be full.

shardtree/src/lib.rs
```diff
@@ -235,17 +235,15 @@ impl<
 
         let (append_result, position, checkpoint_id) =
             if let Some(subtree) = self.store.last_shard().map_err(ShardTreeError::Storage)? {
-                match subtree.max_position() {
-                    // If the subtree is full, then construct a successor tree.
```
str4d, line 238: Keep this comment?

shardtree/src/lib.rs
```diff
@@ -235,17 +235,15 @@ impl<
...
+                if subtree.root().is_full() {
+                    let addr = subtree.root_addr;
+                    if subtree.root_addr.index() < Self::max_subtree_index() {
+                        LocatedTree::empty(addr.next_at_level()).append(value, retention)?
+                    } else {
+                        return Err(InsertionError::TreeFull.into());
                     }
-                    _ => subtree.append(value, retention)?,
+                } else {
+                    subtree.append(value, retention)?
```
str4d, line 246: I _think_ this is well-defined in the context of `!is_definitely_full_on_right()` (nee `!is_full()`) when given the context of this `append()` method (namely that all right-most `Nil` nodes in the last subtree must correspond to actually-absent data, which can only be ensured by the caller).

str4d (changes requested):
Reviewed 68935c8677a8e25dd713d0bae846dd4a1369e926.

str4d:
Also, fix the clippy lints :)

shardtree/src/tree.rs
nuttycom: I don't think that `definitely` adds anything. Maybe it would be better to name this `can_append` and invert the sense.

nuttycom: I had originally documented this as "no new nodes can be appended without introducing a new parent node with this node as the left child" but then decided that that was excessive; *this* tree can't have anything appended to it if it's full.

shardtree/src/legacy.rs
```diff
@@ -203,6 +203,7 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
                         marking: Marking::None,
                     },
                     self.root_addr.level(),
+                    true,
```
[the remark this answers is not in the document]
nuttycom: It occurs to me now that, when inserting a frontier into a tree, what should actually happen is that the observed-ness of the root node being replaced should be propagated downward; this works better than the caller of `from_frontier_parts` having to decide what's appropriate (because they may not be able to know) and this then works well with the update to use a `Parent` with `Nil` children at the root of each shard; once it's been fully scanned and turned into a `Leaf`, subsequent `Frontier` insertions should also use `Leaf` nodes for the inserted ommers.

shardtree/src/prunable.rs
```diff
@@ -793,8 +795,19 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+                subtree = Tree::parent(
+                    None,
+                    // For the leaf level of the tree, or if all ommers are fully observed,
+                    // return a leaf; otherwise return an annotated empty parent node to
+                    // signify that the value of that subtree root has not actually been
+                    // observed.
+                    if ommers_fully_observed || addr.level() == Level::ZERO {
+                        Tree::leaf((left, RetentionFlags::EPHEMERAL))
+                    } else {
+                        Tree::empty_annotated(Some(Arc::new(left)))
```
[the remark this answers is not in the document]
nuttycom: As discussed in pairing, the caller can't know whether ommers_fully_observed is true or not, so we're going back to the representation that uses `Leaf` for all terminal nodes for which the value is known.

str4d (approved):
utACK 192cb3ffda7b5fa018d8efa89551a7376275e4f8

PR #111, opened 2024-09-06
Add max_checkpoints getter on ShardTree

ec2:
I'm currently implementing (de)serialization on the ShardTree and need this to perfectly roundtrip it. Regardless, don't see why this getter shouldn't be exposed anyways.


nuttycom (approved):
ACK, looks good to me.

PR #112, opened 2024-09-10
Immutable For Each on ShardStore Checkpoints

ec2:
The only ways to iterate over Checkpoints in the ShardStore is
1. Use the `with_checkpoints` method. The problem with this is that it require a mut reference to the store which isn't always available (e.g. in serialization).
2. Get the `min_checkpoint_id` and the `max_checkpoint_id` and then casting that to some primitive type if `CheckpointId` is `From/To <numeric>` then calling `get_checkpoint` over the range of that (very not ergonomic!!!).

This PR introduces `for_each_checkpoint` which is literally the same as `with_checkpoints` but takes an immutable reference to the store. 

I actually wanted to rename `with_checkpoints` to `for_each_checkpoint_mut` and call this `for_each_checkpoint` OR call them `with_checkpoints` and `with_checkpoints_mut` but I didn't want to make a breaking API change!

### DIFF
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to Rust's notion of
 
 ## Unreleased
 
+### Added
+- `shardtree::store::ShardStore::for_each_checkpoint`
+
 ## [0.4.0] - 2024-08-12
 
 This is a bugfix release that fixes a couple of subtle problems related to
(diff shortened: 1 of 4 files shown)


ec2:
> utACK [16eff25](https://github.com/zcash/incrementalmerkletree/commit/16eff253ad2575d48feec04f7387e6507a7dd698) for functionality, this can be merged once a CHANGELOG entry has been added.

CHANGELOG entry added

nuttycom (approved):
utACK

PR #113, opened 2024-09-25
Extract `incrementalmerkletree::testing` module to a separate `incrementalmerkletree-testing` crate.

nuttycom:
This change will allow us to continue to move `incrementalmerkletree` and `shardtree` development forward without having to continue to maintain `bridgetree` compatibility with `incrementalmerkletree::testing` after the next `incrementalmerkletree` release.


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.60.0"
+channel = "1.70.0"
```
str4d, line 2: This changes alters the Rust version that CI checks MSRV against. Per https://github.com/zcash/librustzcash/issues/759 we need at least 1.64 use workspaces, but I don't recall if using workspaces prevents the crate from being depended on downstream with a lower MSRV. How is the `Cargo.toml` adjusted during publishing for workspace dependencies?

shardtree/src/lib.rs
```diff
@@ -1462,9 +1461,7 @@ mod tests {
             // boundary.
             tree.batch_insert(
                 frontier_end + 1,
-                ('g'..='j')
-                    .into_iter()
-                    .map(|c| (c.to_string(), Retention::Ephemeral)),
+                ('g'..='j').map(|c| (c.to_string(), Retention::Ephemeral)),
```
str4d, line 1464: I think these changes in the second commit are silently adapting to the silent MSRV bump done in the first commit, specifically to clippy lint changes. I do not know whether these changes are actually compatible with an MSRV of 1.60; this needs testing.

str4d (commented):
Partial review as of d3db03fe67066df02cb727bc465743c72395c0bf (I'm still checking the large move).

If it turns out that the workspace change does require downstreams to raise MSRV, I would prefer that we explicitly raise MSRV in its own commit instead of implicitly doing.

I'd mildly perfer 1.64 instead of 1.70 for this PR (as that is the only necessary bump; it could be raised again in a subsequent PR as new features are needed).

rust-toolchain.toml
nuttycom: I'll try 1.64

str4d (commented):
Reviewed d3db03fe67066df02cb727bc465743c72395c0bf

rust-toolchain.toml
nuttycom: From what I can tell, I can't build even the individual crates without 1.64 when using common workspace stanzas. I think that 1.64 is an acceptable lower bound on MSRV though.

str4d (approved):
utACK f6a5ae46e36a55395b44f7372698b65f916b72b5

str4d (approved):
Re-utACK d663a8f64f04832ebb3b9dfac868ff3ae7b805de

PR #114, opened 2024-09-25
Release `bridgetree` version 0.6.0.

nuttycom:
This change removes `bridgetree` from the workspace, so that it will no longer be maintained in lockstep with changes to the `incrementalmerkletree` crate. `bridgetree` will only receive bug-fix changes in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,10 @@
 members = [
     "incrementalmerkletree",
     "incrementalmerkletree-testing",
-    "bridgetree",
     "shardtree",
+]
+exclude = [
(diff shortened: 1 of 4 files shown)


daira (approved):
This probably needs a changelog entry. Otherwise utACK.

nuttycom:
@daira there is an appropriate changelog entry for `bridgetree`, and otherwise this is just about how the workspace is organized locally; there's no real change to document from a public CHANGELOG perspective.

str4d (approved):
utACK 13a52fcea4646adeb40beed027c5633f78e687a0

PR #115, opened 2024-09-26
shardtree: Significantly rework handling of checkpoint depths and truncation operations.

nuttycom:


incrementalmerkletree-testing/src/lib.rs
```diff
@@ -34,11 +34,13 @@ pub trait Tree<H, C> {
...
-    /// Obtains the root of the Merkle tree at the specified checkpoint depth
-    /// by hashing against empty nodes up to the maximum height of the tree.
-    /// Returns `None` if there are not enough checkpoints available to reach the
-    /// requested checkpoint depth.
-    fn root(&self, checkpoint_depth: usize) -> Option<H>;
+    /// Obtains the root of the Merkle tree at the specified checkpoint depth by hashing against
+    /// empty nodes up to the maximum height of the tree. If the provided checkpoint depth is
+    /// `None`, the root is computed for all leaves in the tree.
+    ///
+    /// Returns `None` if a checkpoint depth is provided but if there are not enough checkpoints
```
str4d, line 41:
```suggestion
    /// Returns `None` if a checkpoint depth is provided but there are not enough checkpoints
```

incrementalmerkletree-testing/CHANGELOG.md
```diff
@@ -7,5 +7,34 @@ and this project adheres to Rust's notion of
...
 
+This release includes a significant refactoring and rework of several methods
+of the `incrementalmerkletree_testing::Tree` trait. Please read the notes for
+this release carefully as the semantics of important methods have changed.
+These changes may require changes to example tests that rely on this crate; in
+particular, additional checkpoints may be required in circumstances where
+rewind operations are being applied.
+
+### Added
+- `incrementalmerkletree_testing::Tree::checkpoint_count`
```
str4d, line 18: Nit: this is a change to a trait, as it doesn't have a provided implementation.

incrementalmerkletree-testing/src/lib.rs
```diff
@@ -34,11 +34,13 @@ pub trait Tree<H, C> {
     /// Return a set of all the positions for which we have marked.
     fn marked_positions(&self) -> BTreeSet<Position>;
 
-    /// Obtains the root of the Merkle tree at the specified checkpoint depth
-    /// by hashing against empty nodes up to the maximum height of the tree.
-    /// Returns `None` if there are not enough checkpoints available to reach the
-    /// requested checkpoint depth.
-    fn root(&self, checkpoint_depth: usize) -> Option<H>;
+    /// Obtains the root of the Merkle tree at the specified checkpoint depth by hashing against
```
str4d, line 37: The trait uses the word "depth" to mean two things:
- The vertical distance below the root that the tree extends.
- The horizontal distance to the right of the most recent checkpoint through the tree's checkpoints.

It would be nice to find an alternative word for the latter, or clarify what "checkpoint depth" means in the few places it gets used.

incrementalmerkletree-testing/src/lib.rs
```diff
@@ -57,12 +59,19 @@ pub trait Tree<H, C> {
     /// less than or equal to the maximum checkpoint identifier observed.
     fn checkpoint(&mut self, id: C) -> bool;
 
-    /// Rewinds the tree state to the previous checkpoint, and then removes that checkpoint record.
+    /// Returns the number of checkpoints in the tree.
+    fn checkpoint_count(&self) -> usize;
+
+    /// Rewinds the tree state to the checkpoint at the specified depth.
```
str4d, line 65: This is what triggered my prior comment: "at the specified depth" sounds like a vertical measure (e.g. at the root of some subtree below the tree root) rather than the intended horizontal measure.

incrementalmerkletree-testing/src/complete_tree.rs
```diff
@@ -299,14 +297,35 @@ impl<H: Hashable + PartialEq + Clone, C: Ord + Clone + core::fmt::Debug, const D
...
-        if let Some((id, c)) = self.checkpoints.iter().rev().next() {
-            self.leaves.truncate(c.leaves_len);
-            for pos in c.marked.iter() {
-                self.marks.remove(pos);
+    fn checkpoint_count(&self) -> usize {
+        self.checkpoints.len()
+    }
+
+    fn rewind(&mut self, depth: usize) -> bool {
+        if self.checkpoints.len() > depth {
```
daira, line 305: Shouldn't it be valid to rewind by exactly the number of checkpoints?

incrementalmerkletree-testing/src/complete_tree.rs
```diff
@@ -237,21 +236,20 @@ impl<H: Hashable + PartialEq + Clone, C: Ord + Clone + core::fmt::Debug, const D
         }
     }
 
-    fn root(&self, checkpoint_depth: usize) -> Option<H> {
-        self.leaves_at_checkpoint_depth(checkpoint_depth)
-            .and_then(|len| root(&self.leaves[0..len], DEPTH))
+    fn root(&self, checkpoint_depth: Option<usize>) -> Option<H> {
```
daira, line 239: Document this.

incrementalmerkletree-testing/src/lib.rs
```diff
@@ -57,12 +59,19 @@ pub trait Tree<H, C> {
...
+    ///
+    /// A depth value of `0` removes all data added to the tree after the most recently-added
+    /// checkpoint. The checkpoint at the specified depth is retained, but all data and metadata
+    /// related to operations on the tree that occurred after the checkpoint was created is
+    /// discarded.
```
str4d, line 70: The first and second sentences seem (to me) to duplicate information, but it's past midnight and I can't think of better wording.

incrementalmerkletree-testing/src/complete_tree.rs
nuttycom: Depth is treated as a (reverse-ordered) 0-based index, so it must be less than the length.

incrementalmerkletree-testing/src/complete_tree.rs
```diff
@@ -237,21 +236,20 @@ impl<H: Hashable + PartialEq + Clone, C: Ord + Clone + core::fmt::Debug, const D
...
+        if self.marks.contains(&position) {
             let leaves_len = self.leaves_at_checkpoint_depth(checkpoint_depth)?;
-            let c_idx = self.checkpoints.len() - checkpoint_depth;
-            if self
-                .checkpoints
-                .iter()
-                .skip(c_idx)
-                .any(|(_, c)| c.marked.contains(&position))
-            {
+            if u64::from(position) >= u64::try_from(leaves_len).unwrap() {
```
str4d, line 252: This unwrap seems to not match the documentation for `Tree::witness`; if `leaves_len.is_none()` then "no checkpoint is available at the specified depth". If this is now a case that `Tree` impls should panic, update the trait method documentation.

incrementalmerkletree-testing/src/complete_tree.rs
```diff
@@ -299,14 +297,35 @@ impl<H: Hashable + PartialEq + Clone, C: Ord + Clone + core::fmt::Debug, const D
...
+            let mut to_delete = vec![];
+            for (idx, (id, c)) in self
+                .checkpoints
+                .iter_mut()
+                .rev()
+                .enumerate()
+                .take(depth + 1)
+            {
+                for pos in c.marked.iter() {
+                    self.marks.remove(pos);
```
str4d, line 315: To clarify, this is correct because the checkpoint's `marked` field stores changes _after_ that checkpoint? (The checkpoint is acting here as the potato box.)

str4d (commented):
Reviewed 0c01af944a3b0b7a0c9c4e4ffbd1b2aacba3fc9f (the first commit), flushing comments.

incrementalmerkletree-testing/src/complete_tree.rs
nuttycom: That unwrap is for the `u64::try_from`.

incrementalmerkletree-testing/src/complete_tree.rs
nuttycom: Correct.

shardtree/src/lib.rs
```diff
@@ -1155,26 +1156,21 @@ impl<
     /// Computes the witness for the leaf at the specified position, as of the given checkpoint
     /// depth.
     ///
-    /// Returns the witness as of the most recently appended leaf if `checkpoint_depth == 0`. Note
-    /// that if the most recently appended leaf is also a checkpoint, this will return the same
-    /// result as `checkpoint_depth == 1`.
+    /// Returns `ShardTreeError::Query(QueryError::CheckpointPruned)` if no checkpoint exists at
+    /// the requested depth.
     pub fn witness_at_checkpoint_depth(
```
str4d, line 1161: AFAICT this change (specifically the 1-indexing to 0-indexing) is not documented in the changelog. Unlike `witness_at_checkpoint_id` there's no method API change because this method returns an error instead of `Ok(None)` if the checkpoint does not exist.

The API difference separately seems like an API consistency bug? If it is intentional, document the difference (and when the distinction should be made use of by downstream crate users) in both methods' documentation (cross-referencing each other), for both pairs (caching and non).

shardtree/src/lib.rs
```diff
@@ -1183,59 +1179,80 @@ impl<
     /// This implementation will mutate the tree to cache intermediate root (ommer) values that are
     /// computed in the process of constructing the witness, so as to avoid the need to recompute
     /// those values from potentially large numbers of subtree roots in the future.
+    ///
+    /// Returns `ShardTreeError::Query(QueryError::CheckpointPruned)` if no checkpoint exists at
+    /// the requested depth.
     pub fn witness_at_checkpoint_depth_caching(
         &mut self,
         position: Position,
         checkpoint_depth: usize,
```
str4d, line 1188: Ditto: the change of this method from 1-indexing to 0-indexing is not documented in the changelog.

str4d (changes requested):
Reviewed 818941800e7553fbc86c8af1f7a31df20e18c695.

incrementalmerkletree-testing/src/complete_tree.rs
str4d: Ah, I missed the trailing `?` on the `self.leaves_at_checkpoint_depth()` call.

shardtree/src/lib.rs
nuttycom: The API difference here was intentional, but I could be convinced to make the two consistent. My rationale was that requesting a specific checkpoint identifier is akin to requesting a specific map key, whereas requesting a witness using a checkpoint depth can either fail because no checkpoint is available, or because the checkpoint is for the empty tree (and consequently a request for a witness as of that tree state is nonsensical.) However, maybe those two kinds of errors should not be conflated.

str4d: The depth query is kind of like an index query into a `Vec`, which for things like `Vec::get` return `None` on out-of-bounds queries. But in that case, querying an empty `Vec` is semantically equivalent to querying past the end of a non-empty `Vec`, whereas here I can kinda see the benefit of distinguishing "there is a witness at this depth, we just don't have information for it" vs "there is no possible witness at this depth" (i.e. any `depth >= tree_size`).

nuttycom:
[force-pushed](https://github.com/zcash/incrementalmerkletree/compare/818941800e7553fbc86c8af1f7a31df20e18c695..9a77e51cc4c99feb6af3bd4146e505e2c70b6677) to address comments from code review.

incrementalmerkletree-testing/src/lib.rs
```diff
@@ -13,56 +13,84 @@ pub mod complete_tree;
...
 pub trait Tree<H, C> {
-    /// Returns the depth of the tree.
+    /// Returns the number of levels in the tree.
     fn depth(&self) -> u8;
 
-    /// Appends a new value to the tree at the next available slot.
-    /// Returns true if successful and false if the tree would exceed
-    /// the maximum allowed depth.
+    /// Appends a new value to the tree at the next available slot. Returns true if successful and
+    /// false if the tree would exceed the maximum allowed number of levels in the tree.
```
str4d, line 23:
```suggestion
    /// Appends a new value to the tree at the next available slot.
    ///
    /// Returns true if successful and false if the tree would exceed
    /// the maximum allowed number of levels in the tree.
```

shardtree/CHANGELOG.md
```diff
@@ -7,11 +7,90 @@ and this project adheres to Rust's notion of
...
+    changed such that if a checkpoint depth is provided, it is now treated
+    strictly as a zero-based index into the checkpoints of the tree in reverse
+    order of checkpoint identifier, and an error is returned if no checkpoint
+    exists at the given index. A `None` value passed for this argument causes
+    it to return the maximum position among all leaves added to the tree.
+  - `ShardTree::root_at_checkpoint_id` and `root_at_checkpoint_id_caching` now
+    each return the computed root as an optional value, returning `Ok(None)` if
+    the checkpoint corresponding to the requested identifier does not exist.
+  - `ShardTree::root_at_checkpoint_depth` and `root_at_checkpoint_depth_caching`
+    now takes the `checkpoint_depth` argument as `Option<usize>` instead of
```
str4d, line 55:
```suggestion
    now take their `checkpoint_depth` argument as `Option<usize>` instead of
```
or "now each take" if you want to match the language from the previous bulletpoint.

shardtree/CHANGELOG.md
```diff
@@ -7,11 +7,90 @@ and this project adheres to Rust's notion of
...
+    `usize`. The semantics of this method are now changed such that if a
```
str4d, line 56:
```suggestion
    `usize`. The semantics of these methods are now changed such that if a
```

shardtree/CHANGELOG.md
```diff
@@ -7,11 +7,90 @@ and this project adheres to Rust's notion of
...
+    computed over all of the leaves in the tree. These methods now each return
+    the computed root as an optional value, returning `Ok(None)` if no checkpoint
+    exist at the requested checkpoint depth.
+  - `ShardTree::witness_at_checkpoint_id` and `witness_at_checkpoint_id_caching`
+    now each return the computed witness as an optional value, returning
+    `Ok(None)` if the checkpoint corresponding to the requested identifier does
+    not exist.
+  - `ShardTree::witness_at_checkpoint_depth` and `witness_at_checkpoint_depth_caching`
+    now each return the computed witness as an optional value, returning
+    `Ok(None)` if no checkpoint was available at the given checkpoint depth.The
```
str4d, line 70:
```suggestion
    `Ok(None)` if no checkpoint was available at the given checkpoint depth. The
```

shardtree/CHANGELOG.md
```diff
@@ -7,11 +7,90 @@ and this project adheres to Rust's notion of
...
+    semantics of this method are now changed such that if a checkpoint depth is
+    provided, it is now treated strictly as a zero-based index into the
+    checkpoints of the tree in reverse order of checkpoint identifier, and an
+    error is returned if no checkpoint exists at the given index. IT IS NO LONGER
```
str4d, line 74: This doesn't match the previous sentence, which says that `Ok(None)` is returned if no checkpoint was available at the given checkpoint depth. (Also they disagree on "depth" vs "index".)

str4d (approved):
utACK 9a77e51cc4c99feb6af3bd4146e505e2c70b6677 after doc fixes

shardtree/CHANGELOG.md
nuttycom, line 74:
```suggestion
    checkpoints of the tree in reverse order of checkpoint identifier. IT IS NO LONGER
```

str4d (approved):
Re-utACK 3f59900acf460af3e08a6ec503adf907ed9bb63b

PR #116, opened 2024-10-02
Use the stable incrementalmerkletree 0.7 dependency in `incrementalmerkletree-testing` and `shardtree`

nuttycom:


str4d (approved):
ACK 518fa387d7041370e6e20ed0397d3efe7703dff9. It would be nice to not need to make developing in this repo harder, but as long as we aren't making frequent changes to `incrementalmerkletree` and instead are just changing `shardtree`, this makes downstream consumption easier.

daira (approved):
utACK

Issue #118, opened 2024-11-22
Audit uses of `.unwrap`/`.expect` in shardtree


daira:
> For the unwrap calls, i see that :
> * [batch.rs](http://batch.rs/) has 1 non-test unwrap
> * [legacy.rs](http://legacy.rs/) has 1 non-test unwrap
> * [lib.rs](http://lib.rs/) has 3 non-test unwrap
> * [prunable.rs](http://prunable.rs/) has 8 non-test  unwrap calls
> * [caching.rs](http://caching.rs/) has 15 non-test  unwrap calls, that are related to the DB access which could be unstable(for ex. due disk space issues)

At least the unwraps for DB access seem like a bug to me.

str4d:
> > * `batch.rs` has 1 non-test unwrap

This (along with one unwrap in `prunable.rs`) is due to `BatchInsertionResult.max_insert_position` being an `Option<Position>`, but I cannot find anywhere this gets set to `None`.

> > * `legacy.rs` has 1 non-test unwrap

This should be justifiable because of the invariant that you cannot construct a witness for an empty tree. However, `incrementalmerkletree::witness::IncrementalWitness` does permit construction from an empty tree, so we can't currently rely on that invariant. We should fix `incrementalmerkletree`.

> > * `lib.rs` has 3 non-test unwrap
> > * `caching.rs` has 15 non-test  unwrap calls, that are related to the DB access which could be unstable(for ex. due disk space issues)
> 
> At least the unwraps for DB access seem like a bug to me.

These are all justifiable.

> > * `prunable.rs` has 8 non-test  unwrap calls

All except two of these (exluding the one mentioned above) are justifiable. The other two are technically justifiable in the context they get used, but their public APIs don't enforce the necessary invariant, so should be fixed (with the `unwrap` moving out into the place these APIs get used where the invariant can be justified).

str4d:
> > > * `batch.rs` has 1 non-test unwrap
> 
> This (along with one unwrap in `prunable.rs`) is due to `BatchInsertionResult.max_insert_position` being an `Option<Position>`, but I cannot find anywhere this gets set to `None`.
> 
> > > * `legacy.rs` has 1 non-test unwrap
> 
> This should be justifiable because of the invariant that you cannot construct a witness for an empty tree. However, `incrementalmerkletree::witness::IncrementalWitness` does permit construction from an empty tree, so we can't currently rely on that invariant. We should fix `incrementalmerkletree`.
> 
[...]

PR #119, opened 2024-11-23
Replace justified `unwraps` in `shardtree` with `expect`s

str4d:
Part of #118.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -1162,7 +1162,8 @@ impl<
             cur_addr = cur_addr.parent();
         }
 
-        Ok(MerklePath::from_parts(witness, position).unwrap())
+        Ok(MerklePath::from_parts(witness, position)
+            .expect("witness has length DEPTH because we extended it to the root"))
     }
 
     fn witness_internal(
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -406,13 +406,16 @@ impl<
 
     /// Adds a checkpoint at the rightmost leaf state of the tree.
     pub fn checkpoint(&mut self, checkpoint_id: C) -> Result<bool, ShardTreeError<S::Error>> {
+        /// Pre-condition: `root_addr` must be the address of `root`.
         fn go<H: Hashable + Clone + PartialEq>(
             root_addr: Address,
             root: &PrunableTree<H>,
         ) -> Option<(PrunableTree<H>, Position)> {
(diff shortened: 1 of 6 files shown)


shardtree/src/lib.rs
```diff
@@ -406,13 +406,16 @@ impl<
 
     /// Adds a checkpoint at the rightmost leaf state of the tree.
     pub fn checkpoint(&mut self, checkpoint_id: C) -> Result<bool, ShardTreeError<S::Error>> {
+        /// Pre-condition: `root_addr` must be the address of `root`.
```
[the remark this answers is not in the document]
str4d: > it is a programming error for these to be inconsistent.

I note that the public API `LocatedTree::from_parts` does not enforce that its arguments are consistent; we could potentially strengthen this by making that method fallible and checking that the actual encoded depth of the provided tree matches the provided root address (which would ensure that all parents within the provided tree correspond to a subtree node of the provided root address that can have children).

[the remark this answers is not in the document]
str4d: I've updated this PR to add the consistency check to `LocatedTree::from_parts`.

shardtree/src/prunable.rs
```diff
@@ -440,8 +446,10 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
     /// Returns either the witness for the leaf at the specified position, or an error that
     /// describes the causes of failure.
     pub fn witness(&self, position: Position, truncate_at: Position) -> Result<Vec<H>, QueryError> {
-        // traverse down to the desired leaf position, and then construct
-        // the authentication path on the way back up.
+        /// Traverse down to the desired leaf position, and then construct
+        /// the authentication path on the way back up.
+        //
```
daira, line 451: Non-blocking nit:
```suggestion
        ///
```

daira (approved):
utACK

shardtree/src/prunable.rs
nuttycom: I didn't think that `///` comments were allowed on internal documentation? I suspect this whole comment will produce a linting or doc error.

nuttycom (approved):
utACK modulo doc/comment linting c05f2fac006b7336939c1854c774f3642f581214

shardtree/src/prunable.rs
nuttycom: I verified that `cargo doc` accepts these without complaint.

PR #120, opened 2024-11-23
incrementalmerkletree: Check `IncrementalWitness` at construction

str4d:
A witness cannot exist for an empty tree.

Part of zcash/incrementalmerkletree#118.


daira (approved):
utACK

PR #121, opened 2024-12-02
Backport #119 to shardtree v0.3.x

nuttycom:
Backport of #119

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -392,13 +392,16 @@ impl<
 
     /// Adds a checkpoint at the rightmost leaf state of the tree.
     pub fn checkpoint(&mut self, checkpoint_id: C) -> Result<bool, ShardTreeError<S::Error>> {
+        /// Pre-condition: `root_addr` must be the address of `root`.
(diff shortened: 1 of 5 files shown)


daira (approved):
utACK. (I think @str4d might have said he had another change to make.)

shardtree/src/lib.rs
```diff
@@ -1103,7 +1109,8 @@ impl<
             cur_addr = cur_addr.parent();
         }
 
-        Ok(MerklePath::from_parts(witness, position).unwrap())
+        Ok(MerklePath::from_parts(witness, position)
+            .expect("witness has length DEPTH because we extended it to the root"))
```
daira, line 1113: That depends on collision resistance, but I'm fine with using `.expect` in that case.

daira (approved):
utACK 392e395fdedebcb6cbfbae33d445f1aa2b36f482

str4d (approved):
utACK 77fa60088fd7bd65e9c4af73864dc67d6e52ae9b. `unwrap`s now all document why they are there. Several of them can be reachable if the caller makes a programming error; this can only be fixed with a subsequent breaking API change (as in my other PRs).

PR #124, opened 2024-12-12
Merge `shardtree` hotfixes back to `main` & update `shardtree` to `incrementalmerkletree 0.8.0`

nuttycom:


shardtree/src/tree.rs
```diff
@@ -199,33 +199,35 @@ pub struct LocatedTree<A, V> {
 impl<A, V> LocatedTree<A, V> {
     /// Constructs a new LocatedTree from its constituent parts.
     ///
-    /// Returns `None` if `root_addr` is inconsistent with `root` (in particular, if the
-    /// level of `root_addr` is too small to contain `tree`).
-    pub fn from_parts(root_addr: Address, root: Tree<A, V>) -> Option<Self> {
+    /// Returns the newly constructed error, or the address at which the provided tree extends
+    /// beyond the position range of the provided root address.
+    pub fn from_parts(root_addr: Address, root: Tree<A, V>) -> Result<Self, Address> {
```
nuttycom, line 204: I decided to change this to a `Result` instead of an `Option` because there are places in the `ShardStore` API that return `Option<LocatedTree>`, and I do not want inconsistency to be able to be confused with malformedness.

daira (approved):
utACK

PR #127, opened 2025-01-28
Release shardtree version 0.6

nuttycom:


shardtree/CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to Rust's notion of
 
 ## Unreleased
 
+## [0.6.0] - 2025-01-28
+
 ### Changed
+- Update to `incrementalmerkletree 0.8.1`
```
daira, line 13: https://github.com/zcash/incrementalmerkletree/blob/3b21d2123edde31d6abfbce19db77fe923cdfaab/Cargo.toml#L21 says the workspace depends on incrementalmerkletree 0.8.

daira (approved):
utACK

shardtree/CHANGELOG.md
str4d: Yep, this should just be
```suggestion
- Update to `incrementalmerkletree 0.8`
```
unless the intention is specifically to pin `0.8.1`, which I don't think is actually necessary.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ categories = ["algorithms", "data-structures"]
 
 [workspace.dependencies]
 # Intra-workspace dependencies
-incrementalmerkletree = { version = "0.8", path = "incrementalmerkletree" }
+incrementalmerkletree = { version = "0.8.1", path = "incrementalmerkletree" }
```
str4d, line 21: As noted in my other comment, I don't think this is actually necessary; `cargo` will consider `0.8.0` and `0.8.1` as compatible, and if you're not in a `no-std` environment then `0.8.1` contains nothing that you need (so this is just an overconstraint).

If, by contrast, the being-published `shardtree` depended on a new API exposed by `incrementalmerkletree` then this would be the correct way to express the dependency.

str4d (approved):
utACK 04280c42ab822feec58ac460290d1c3149774f67 with nit

PR #128, opened 2025-01-30
Release incrementalmerkletree-testing 0.3

nuttycom:


daira (approved):
utACK

incrementalmerkletree-testing/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## Unreleased
 
+## [0.3.0] - 2025-01-30
+
+### Changed
+- Updated to `incrementalmerkletree 0.8`
```
str4d, line 13: Technically due to how the versions are set in the root `Cargo.toml` this is updating to `0.8.1` specifically, but meh I don't think anyone else will notice in this case.

str4d (approved):
utACK 44065917a399a921a2151bd14d399a192eb61960

PR #131, opened 2025-02-17
shardtree: make construction of fully-empty Parent nodes an error

nuttycom:
Ideally the errors here would all be panics, as construction of such a node represents a programming error; however, it is preferable to return extra contextual information about the circumstance that led to this error where possible, so we model this as an error where possible without altering the public API.


shardtree/src/batch.rs
```diff
@@ -238,7 +239,9 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
             let minimal_tree_addr =
                 Address::from(position_range.start).common_ancestor(&last_position.into());
             trace!("Building minimal tree at {:?}", minimal_tree_addr);
-            build_minimal_tree(fragments, minimal_tree_addr, prune_below).map(
+            build_minimal_tree(fragments, minimal_tree_addr, prune_below)
+                .expect("construction of minimal tree does not result in creation of invalid parent nodes")
+                .map(
                 |(to_insert, contains_marked, incomplete)| BatchInsertionResult {
```
daira, line 245: I could be wrong, but this doesn't look rustfmt'd.

shardtree/src/batch.rs
```diff
@@ -286,7 +291,7 @@ fn combine_with_empty<H: Hashable + Clone + PartialEq>(
     incomplete: &mut Vec<IncompleteAt>,
     contains_marked: bool,
     prune_below: Level,
-) -> LocatedPrunableTree<H> {
+) -> Result<LocatedPrunableTree<H>, Address> {
```
daira, line 294: Using an `Address` directly as an error type is odd.

shardtree/src/batch.rs
```diff
@@ -313,27 +318,27 @@ fn build_minimal_tree<H: Hashable + Clone + PartialEq>(
     mut xs: Vec<(LocatedPrunableTree<H>, bool)>,
     root_addr: Address,
     prune_below: Level,
-) -> Option<(LocatedPrunableTree<H>, bool, Vec<IncompleteAt>)> {
+) -> Result<Option<(LocatedPrunableTree<H>, bool, Vec<IncompleteAt>)>, Address> {
```
daira, line 321: This doesn't look like a "very complex type" to me; maybe the clippy warning is over-sensitive and its sensitivity should be turned down globally.

daira (changes requested):
Blocking comments/questions.

shardtree/src/batch.rs
nuttycom: It's the address at which inconsistencies were found; anyway, Result is just a badly-named Either.

shardtree/src/prunable.rs
```diff
@@ -586,39 +600,84 @@ impl<H: Hashable + Clone + PartialEq> LocatedPrunableTree<H> {
...
+    // and recording the presence of those incomplete nodes when necessary
+    fn extend_to_level(
+        self,
+        level: Level,
+        ann: Option<Arc<H>>,
+    ) -> Result<(PrunableTree<H>, Vec<Address>), Address> {
+        // construct the replacement node bottom-up
+        let mut node = self;
+        let mut incomplete = vec![];
+        while node.root_addr.level() < level {
```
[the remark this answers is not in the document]
nuttycom: Ah, yeah, factoring this out as a standalone function instead of a closure means that the invariants that were enforced by the enclosing scope are no longer checked. I don't think that's a blocker though; this is just used in the one place for now.

shardtree/src/batch.rs
daira: We've discussed this before I think. This is not an idiomatic use of `Result`: it normally takes an error type on the right. The error type can just be a wrapper around `Address` if you want, but by convention it implements `Debug`, `fmt::Display`, and `std::error::Error` if appropriate, something like this:

```rust
/// Error indicating an inconsistency at a given address.
#[derive(Debug)]
pub struct InconsistencyError(pub Address);

impl fmt::Display for InconsistencyError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Inconsistency found at address {}", self.0)
    }
}

#[cfg(feature = "std")]
impl std::error::Error for InconsistencyError {}
```

daira (approved):
I would prefer that a specific error type be used instead of `Address`, but I'm not insisting on it. utACK

shardtree/src/batch.rs
nuttycom: This is an odd rustfmt inconsistency: I discovered that if I make the string on line 243 shorter, then this formats more sensibly.

shardtree/src/batch.rs
daira: Okay that's fine by me.

PR #139, opened 2026-02-12
Add methods to obtain the current frontier for `ShardTree` and `LocatedPrunableTree` 

nuttycom:
Blocks zcash/librustzcash#2121

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
 exclude = [
     "bridgetree",
 ]
+resolver = "2"
 
 [workspace.package]
 edition = "2021"
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -160,6 +160,42 @@ impl<
         Ok(result)
     }
 
+    /// Returns the frontier of the tree.
+    ///
+    /// The frontier consists of the rightmost leaf and the ommers (sibling hashes)
+    /// needed to compute the root. This method assembles ommers from within the
+    /// last shard (below `SHARD_HEIGHT`) and from sibling subtree roots above the
+    /// shard level.
+    ///
+    /// Returns [`Frontier::empty`] if the tree has no shards or if the frontier
+    /// cannot be determined from the available data.
+    pub fn frontier(&self) -> Result<Frontier<H, DEPTH>, ShardTreeError<S::Error>> {
+        let last_shard = self.store.last_shard().map_err(ShardTreeError::Storage)?;
+        let last_shard = match last_shard {
+            Some(s) => s,
+            None => return Ok(Frontier::empty()),
+        };
+
+        let (position, leaf, mut ommers) = match last_shard.frontier_ommers() {
+            Some(parts) => parts,
+            None => return Ok(Frontier::empty()),
+        };
+
+        // Walk from the shard root address up to the tree root, collecting ommers
+        // from sibling subtrees above the shard level.
+        let mut cur_addr = last_shard.root_addr;
+        while cur_addr.level() < Level::from(DEPTH) {
+            if cur_addr.is_right_child() {
+                let sibling_hash = self.root(cur_addr.sibling(), position + 1)?;
+                ommers.push(sibling_hash);
+            }
+            cur_addr = cur_addr.parent();
+        }
+
+        Frontier::from_parts(position, leaf, ommers)
+            .map_err(|_| ShardTreeError::Query(QueryError::TreeIncomplete(vec![Self::root_addr()])))
+    }
+
     /// Inserts a new root into the tree at the given address.
     ///
     /// The level associated with the given address may not exceed `DEPTH`.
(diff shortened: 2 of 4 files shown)


shardtree/src/lib.rs
```diff
@@ -160,6 +160,42 @@ impl<
...
+    /// Returns the frontier of the tree.
+    ///
+    /// The frontier consists of the rightmost leaf and the ommers (sibling hashes)
+    /// needed to compute the root. This method assembles ommers from within the
+    /// last shard (below `SHARD_HEIGHT`) and from sibling subtree roots above the
+    /// shard level.
+    ///
+    /// Returns [`Frontier::empty`] if the tree has no shards or if the frontier
+    /// cannot be determined from the available data.
+    pub fn frontier(&self) -> Result<Frontier<H, DEPTH>, ShardTreeError<S::Error>> {
```
str4d: Document this addition in the changelog.

shardtree/src/lib.rs
```diff
@@ -160,6 +160,42 @@ impl<
...
 
+    /// Returns the frontier of the tree.
+    ///
+    /// The frontier consists of the rightmost leaf and the ommers (sibling hashes)
+    /// needed to compute the root. This method assembles ommers from within the
+    /// last shard (below `SHARD_HEIGHT`) and from sibling subtree roots above the
+    /// shard level.
+    ///
+    /// Returns [`Frontier::empty`] if the tree has no shards or if the frontier
+    /// cannot be determined from the available data.
```
str4d: This return value makes sense if the tree has no shards (and thus is actually empty, which can legitimately occur in e.g. the period of time between adding support for a new shielded pool to code and consensus, but before it activates in a NU). But if there is incomplete data it feels like this should be distinguished with e.g. a `None` (which the caller can map to their "wallet needs further syncing" error). I guess the reason you've aliased these two is that the no-shards case could *also* mean "wallet needs further syncing", but I feel like that distinction needs to be made by the caller (i.e. if the caller knows they are fully synced, then no-shards actually does mean "tree empty", whereas insufficient information means data corruption).

shardtree/src/lib.rs
```diff
@@ -1472,6 +1508,108 @@ mod tests {
...
+
+        let mut tree: ShardTree<MemoryShardStore<String, u32>, 4, 3> =
+            ShardTree::new(MemoryShardStore::empty(), 100);
+        tree.insert_frontier_nodes(original.clone(), Retention::Ephemeral)
+            .unwrap();
+
+        let frontier = tree.frontier().unwrap();
+        assert_eq!(frontier.value(), Some(&original));
+    }
+
```
str4d: This test is identical to `frontier_roundtrip()`, and indeed as I was reviewing this test and `frontier_within_single_shard` I noted to myself that these were really just roundtrip tests. EIther figure out what is actually testable separately from a roundtrip test and rework this test to test it, or remove one of the two tests and ensure the other(s) that do rountrip testing are labeled as such.

str4d (changes requested):
Reviewed 7b6263ee2af31b2db0131e28928b83a4d52f4e83. I also confirmed that no breaking changes have been made to `shardtree` in `main`, so we *could* treat `main` as effectively the maintenance branch for both `shardtree` and `incrementalmerkletree`, but we might also want to consider creating explicit maintenance branches for the two (and then merging `incrementalmerkletree`'s into `shardtree`'s when releases occur).

shardtree/src/lib.rs
nuttycom: Incomplete data is indicated by the `QueryError::TreeIncomplete` error. But I need to think about the conditions that arise in line 179; that should maybe also return `QueryError::TreeIncomplete` (or make it possible to distinguish between these two situations.)

nuttycom:
[force-pushed](https://github.com/zcash/incrementalmerkletree/compare/7b6263e..ccae5d75a8095e3399e002055a49ff845890e8ac) to address comments from code review.

shardtree/src/prunable.rs
```diff
@@ -388,6 +388,38 @@ pub struct IncompleteAt {
     pub required_for_witness: bool,
 }
 
+/// Validation errors that can occur during reconstruction of a Merkle frontier from
+/// its constituent parts.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FrontierError {
```
daira, line 395: Should be non-exhaustive.

shardtree/src/prunable.rs
```diff
@@ -1279,6 +1383,234 @@ mod tests {
...
+        };
+
+        // Position 5 (binary 101) needs 2 ommers, but we only have 1 within the shard
+        // (level 0 ommer "a"). The bit at level 2 is set in position 5, requiring an
+        // ommer we don't have. from_parts rejects this.
+        assert_matches!(t.frontier(), Err(FrontierError::PositionMismatch { .. }));
+    }
+
+    #[test]
+    fn frontier_roundtrip_via_insert() {
```
daira, line 1592: This looks as though it could be a proptest to get better coverage. Non-blocking.

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
 exclude = [
     "bridgetree",
 ]
+resolver = "2"
```
daira, line 10: (Infornational) This is not a breaking change because the `resolver` setting of a dependency is ignored (and it did not require us to bump MSRV).

daira (approved):
utACK with suggestions.

str4d (approved):
utACK 4d877a0688236b681d5510cdccf30a54f5187ee3

PR #143, opened 2026-03-27
Simplify root_internal into flat three-phase structure

nuttycom:
Restructure root_internal from three separate node handlers (Parent, Leaf, Nil) with multiple sub-branches each into a flat three-phase flow: (1) fast-path using node_value(), (2) base case at shard level or target address, (3) descent with inline child extraction. This eliminates the trampoline pattern of constructing synthetic Parent nodes to re-enter the function, and deduplicates the child recursion and result-combining logic that was previously written separately for each node type.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -811,168 +811,138 @@ impl<
         // roots.
         truncate_at: Position,
     ) -> Result<(H, Option<PrunableTree<H>>), ShardTreeError<S::Error>> {
-        match &cap.root.0 {
-            Node::Parent { ann, left, right } => {
-                match ann {
-                    Some(cached_root)
-                        if target_addr.contains(&cap.root_addr)
-                            && truncate_at >= cap.root_addr.position_range_end() =>
-                    {
-                        Ok((cached_root.as_ref().clone(), None))
-                    }
-                    _ => {
-                        // Compute the roots of the left and right children and hash them together.
-                        // We skip computation in any subtrees that will not have data included in
-                        // the final result.
-                        let (l_addr, r_addr) = cap
-                            .root_addr
-                            .children()
-                            .expect("has children because we checked `cap.root` is a parent");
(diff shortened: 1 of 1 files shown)


dannywillems (approved):
utACK c396c49cb3c5875fa33bc006a6eb460a89c82452.

As part of this review, I spent time reading the whole crate, and made a few patches for documentation and tests, some already reviewed by @nullcopy.

I won't merge it now though. The method has different cases (Claude gives me 19 in total), and I will first add regression tests for all of them.

I drew most of the cases on a paper, to convince myself of each path, but I would prefer to have actual test vectors, in case we change again later the implementation.

dannywillems:
> utACK [c396c49](https://github.com/zcash/incrementalmerkletree/commit/c396c49cb3c5875fa33bc006a6eb460a89c82452).
> 
> As part of this review, I spent time reading the whole crate, and made a few patches for documentation and tests, some already reviewed by @nullcopy.
> 
> I won't merge it now though. The method has different cases (Claude gives me 19 in total), and I will first add regression tests for all of them.
> 
> I drew most of the cases on a paper, to convince myself of each path, but I would prefer to have actual test vectors, in case we change again later the implementation.

See https://github.com/zcash/incrementalmerkletree/pull/167. I suggest to merge this patch when/if 167 is approved and merged.

dannywillems:
Claude detected more cases uncovered by !167. Adding more.

EDIT: The patch https://github.com/zcash/incrementalmerkletree/pull/167 has been updated, filling the gaps mentioned in this message before the edit.

dannywillems:
> Claude detected more cases uncovered by !167. Adding more.
> 
> EDIT: The patch #167 has been updated, filling the gaps mentioned in this message before the edit.

See https://github.com/zcash/incrementalmerkletree/pull/170 as a follow-up on this.

nuttycom:
Closing in favor of #170 as this fails the tests added in #167

PR #146, opened 2026-06-07
bridgetree: Memoize the root of the current frontier; release bridgetree 0.7.1

daira:
Release `bridgetree 0.7.1`.

### Optimization

`BridgeTree::root(0)` re-folds the current frontier against empty subtree roots up to the full depth of the tree on every call. For consumers that query the root once per tree mutation —e.g. a per-block consistency check where the common case is a block that appends no leaves— this fixed per-call cost dominates.

Cache the depth-0 root and clear it in `append` and `rewind`, the only operations that change the frontier. `checkpoint`, `mark`, `remove_mark`, and `garbage_collect` preserve the frontier and so preserve the cache, so the cache survives across the checkpoint that bounds each block when that block appends no leaves. Repeated root queries between mutations are now O(1).

### Changed

The MSRV is raised from 1.64 to 1.70 in order to use `std::sync::OnceLock`.

### DIFF
--- a/bridgetree/CHANGELOG.md
+++ b/bridgetree/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- `BridgeTree::root(0)` (the root of the current frontier) is now memoized. The
+  cached value is cleared by `append` and `rewind`, the only operations that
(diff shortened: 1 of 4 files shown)


daira:
I actually clicked the Copilot review thing by accident (curses to dark UI patterns). Oh well.

nuttycom (approved):
utACK

PR #154, opened 2026-06-16
Document (Located?Prunable?)Tree

dannywillems:
Even though browsing the code helps, I still need a bit of time to remember again the next day or week when I come back to the code.

### DIFF
--- a/shardtree/src/prunable.rs
+++ b/shardtree/src/prunable.rs
@@ -98,7 +98,27 @@ impl<C> From<Retention<C>> for RetentionFlags {
     }
 }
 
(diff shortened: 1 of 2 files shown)


shardtree/src/prunable.rs
nullcopy, line 1: (nit) typo in commit message: `s/shardtre/shardtree/`

dannywillems: Fixed.

nullcopy (approved):
ACK d9ced1527db5a53bbf178f01ac5607cc91658e7c

PR #155, opened 2026-06-16
shardtree: document the type parameters semantic

dannywillems:
Built on top of #154

### DIFF
--- a/shardtree/src/tree.rs
+++ b/shardtree/src/tree.rs
@@ -6,6 +6,12 @@ use std::sync::Arc;
 use incrementalmerkletree::{Address, Level, Position};
 
 /// A "pattern functor" for a single layer of a binary tree.
+///
+/// # Type parameters
+/// - `C`: the child type (in [`Tree`] this is `Arc<Tree<A, V>>`).
+/// - `A`: internal (`Parent`) node data, i.e. the annotation attached to an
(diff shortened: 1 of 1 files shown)


shardtree/src/tree.rs
```diff
@@ -6,6 +6,12 @@ use std::sync::Arc;
 use incrementalmerkletree::{Address, Level, Position};
 
 /// A "pattern functor" for a single layer of a binary tree.
+///
+/// # Type parameters
+/// - `C`: the child type (in [`Tree`] this is `Arc<Tree<A, V>>`).
+/// - `A`: internal (`Parent`) node data, i.e. the annotation attached to an
```
nullcopy, line 12: My guess is `A` was for "Ancestor", but that's just a guess. This description seems correct to me

nullcopy (dismissed):
ACK fa02cb21a26e50d70bc7475b4a12919ac6d5ab15

dannywillems:
Rebased to fix conflicts with `main`.

shardtree/src/tree.rs
dannywillems: I confirm that `A` is for annotation, the type that the internal nodes contain.

nullcopy (approved):
utACK 1237bb9f1dd90295976697f9fd3e09f3ac837df3

PR #156, opened 2026-06-16
shardtree: improve documentation of method `root`

dannywillems:
The method name is misleading. A method called `root` for a type representing a Merkle tree generally computes the root of the overall tree. However, this method computes a root relative to a position, with possible tweaks to the subtree (truncating leafs, i.e. considering leafs values as null).

Additionnally, this commit adds a `Panics` section for the potential exceptions the method might raise.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -750,15 +750,34 @@ impl<
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


nullcopy (approved):
ACK a3d469b13624ddaf6284da5eef27ff296b243818

PR #157, opened 2026-06-16
Document and increase regression test coverage for checkpoints pruning

dannywillems:


nullcopy (approved):
utACK e43204ef84dc956e7867415ed43a990d58205099

PR #166, opened 2026-06-16
shardtree: complete top-level documentation

dannywillems:
Add an explanation about the general idea of the crate, with its different main concepts (shards and caps), and links to the main structures/types, for a fast visit of the library.


nullcopy (approved):
ACK a48d25546b83ad6e9fc824fb8aab80774d146851

PR #167, opened 2026-06-17
shardtree: case by case testing for root_internal

dannywillems:
The patch https://github.com/zcash/incrementalmerkletree/pull/143 has been approved, see [review](https://github.com/zcash/incrementalmerkletree/pull/143#pullrequestreview-4515026732). However, as mentioned in the comment, adding regression tests in the form of expected value/computed value, for each case, would add more certainty about the refactorization.
This patch aims to fill that gap.

These tests have been generated using Claude Code 4.8 (1M context), with the following process:
- Move on the branch of 143.
- Ask Claude to go through all different conditional branches to list all possible cases.
- Generate individual tests for each cases, with the expected output for !143.
  - See 71e37c06b2874bc7acfaa9140d1d3a1a51c33e95 and ec8b5b76ceae82a6afd48766356a486f38284397.
- Ensure the tests pass for !143.
- Create a new branch from main with these tests.
- Ensure the tests pass for main
- Ask Claude to go through the existing code of `root_internal`, criticize the new tests, and add additional cases, if any.
  - Claude suggested additional checks for the updated_cap value - untested initially, and new tests.
  - See fea20e62f963d2c9f84650b6da9e9beddf31d57c and 21683b08fa518c2d932121ceb597231f990f625a
- Update the tests with these critics.
- Go back on !143, run the tests, ask Claude to fix the code, _without_ changing the tests.

Ended up with this single discrepancy: https://github.com/zcash/incrementalmerkletree/pull/167#discussion_r3428883837. Patch coming.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -1585,30 +1585,38 @@ mod tests {
 
     #[test]
     fn shardtree_insertion() {
-        let tree: ShardTree<MemoryShardStore<String, u32>, 4, 3> =
-            ShardTree::new(MemoryShardStore::empty(), 100);
+        let tree = empty_tree::<String, 4, 3>();
 
         check_shardtree_insertion(tree)
     }
 
     #[test]
     fn shard_sizes() {
-        let tree: ShardTree<MemoryShardStore<String, u32>, 4, 2> =
-            ShardTree::new(MemoryShardStore::empty(), 100);
+        let tree = empty_tree::<String, 4, 2>();
 
         check_shard_sizes(tree)
     }
 
     #[test]
     fn witness_with_pruned_subtrees() {
-        let tree: ShardTree<MemoryShardStore<String, u32>, 6, 3> =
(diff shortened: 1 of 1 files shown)


shardtree/src/lib.rs
```diff
@@ -2232,4 +2232,684 @@ mod tests {
...
+
+        #[test]
+        fn base_leaf_at_target_truncated() {
+            // A Leaf above the shard level, queried at its own address with
+            // truncation. In the pre-#143 code this is the Leaf handler's
+            // recurse-and-reannotate branch; in #143 it is the Phase 2
+            // `root_addr == target && !Parent` clause (which the other `base_*`
+            // tests reach only with a Nil node).
+            //
+            // Divergence tripwire: the two implementations agree on the
```
dannywillems, line 2499: This is a failing test for !143. I will submit a patch on top of !143. In the meantime, I think that getting this patch reviewed and merged could be a good thing, as it is only tests.

nuttycom (approved):
utACK

PR #170, opened 2026-06-17
shardtree: preserve reannotated cap for Leaf at target in root_internal

dannywillems:
This patch has been built on top of https://github.com/zcash/incrementalmerkletree/pull/167 followed by https://github.com/zcash/incrementalmerkletree/pull/143. The only relevant commit for this patch is da08baeebaaac57ba1603981b19edd9bec6523f0.
The reviewer is invited to do the following to convince themselves the fix is necessary.
```
git checkout 4212175b59ae50b39a4c5cfd826802757721645a
cargo test -p shardtree --release
```
The test `base_leaf_at_target_truncated` will fail. However, with
```
git checkout 21683b08fa518c2d932121ceb597231f990f625a
cargo test -p shardtree --release
```
the same test succeeds. As `base_leaf_at_target_truncated` is considered as a regtest, the behavior of `21683b08fa518c2d932121ceb597231f990f625a` is the expected one.

Review and merge both of them in order, before this one.

---

The Phase 2 base case matched any non-Parent node at the target address,
which routed a Leaf-at-target through root_from_shards and discarded its
cached hash under truncation. Restrict the clause to Nil so a Leaf is
expanded by the Phase 3 descent and reannotated with its original hash,
preserving the cached value in the written-back cap. The annotation is the
original un-truncated hash, never the truncated root, so the entry stays
valid for later non-truncated lookups via the Phase 1 fast path.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -913,168 +913,138 @@ impl<
         target_addr: Address,
         truncate_at: Position,
     ) -> Result<(H, Option<PrunableTree<H>>), ShardTreeError<S::Error>> {
-        match &cap.root.0 {
-            Node::Parent { ann, left, right } => {
-                match ann {
-                    Some(cached_root)
-                        if target_addr.contains(&cap.root_addr)
-                            && truncate_at >= cap.root_addr.position_range_end() =>
-                    {
-                        Ok((cached_root.as_ref().clone(), None))
-                    }
-                    _ => {
-                        // Compute the roots of the left and right children and hash them together.
-                        // We skip computation in any subtrees that will not have data included in
-                        // the final result.
-                        let (l_addr, r_addr) = cap
-                            .root_addr
-                            .children()
-                            .expect("has children because we checked `cap.root` is a parent");
-                        let l_result = if r_addr.contains(&target_addr) {
-                            None
-                        } else {
-                            Some(self.root_internal(
-                                &LocatedPrunableTree {
-                                    root_addr: l_addr,
-                                    root: left.as_ref().clone(),
-                                },
-                                if l_addr.contains(&target_addr) {
-                                    target_addr
-                                } else {
-                                    l_addr
-                                },
-                                truncate_at,
-                            )?)
-                        };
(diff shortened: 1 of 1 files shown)


shardtree/src/lib.rs
```diff
@@ -913,168 +913,140 @@ impl<
...
+                    Arc::new(S::H::combine(l_addr.level(), l, r))
+                }),
+            left: new_left.map_or_else(|| orig_left, Arc::new),
+            right: new_right.map_or_else(|| orig_right, Arc::new),
+        });
+
+        // If the original node was a Leaf, preserve its hash as the Parent annotation
+        // so that future non-truncated lookups can use it via the fast-path.
+        let replacement = match orig_leaf_value {
+            Some(h) => new_parent.reannotate_root(Some(Arc::new(h))),
```
nuttycom, line 1045: Under what circumstance should this differ from that computed for `new_parent` already? Shouldn't it be an error if that conflicts with what we just computed, if it's non-None? Instead of `new_parent.reannotate_root`, should the above construction of `new_parent` check that they're equal if they're both `Some`?

shardtree/src/lib.rs
```diff
@@ -913,168 +913,140 @@ impl<
...
+        // preserving cached values when a Leaf is expanded.
+        let orig_leaf_value = match &cap.root.0 {
+            Node::Leaf { value } => Some(value.0.clone()),
+            _ => None,
+        };
+
+        // Get children: real children for Parent nodes, empty children for Leaf/Nil.
+        let (orig_left, orig_right) = match &cap.root.0 {
+            Node::Parent { left, right, .. } => (left.clone(), right.clone()),
+            _ => (Arc::new(Tree::empty()), Arc::new(Tree::empty())),
```
nuttycom, line 965: What do we actually obtain in the non-parent case here? We can't recurse into `Tree::empty`, so can't we skip the recursive call in these cases? Or does doing it this way just mean fewer special cases?

nuttycom (approved):
utACK 9096a8d61ae0b8ac7d38d2f40c21378d8b54de2f

Issue #173, opened 2026-06-22
Scale the in-memory overlay path: sparse `MemoryShardStore` + dirty-delta `CachingShardStore`


LukasKorba:
## Summary

`CachingShardStore` is already the "write-back overlay" pattern — cache a tree in memory, flush to a
backend — but every operation is whole-tree:

- `load()` eagerly copies **every** shard and **every** checkpoint out of the backend.
- `flush()` re-writes **every** shard and re-adds **every** checkpoint regardless of what changed, and
  it consumes `self` (one-shot, can't flush-and-continue).

In the Zashi/Zodl wallet sync engine ("Slipstream") we built and shipped an optimized version of exactly
this overlay — **sparse seed + dirty-delta flush + incremental (reusable) flush** — as part of a ~12×
wallet-sync speedup, gated by a byte-identical `data.db` oracle. We'd like to fold those wins back
upstream so every consumer benefits — notably `zcash_client_memory`/WebZjs, where the per-flush cost and
in-memory footprint matter most.

This is **two largely independent improvements plus one design question** we'd like your steer on before
we open PRs. We have a working reference implementation for all of it and are offering to do the PRs.

## 1. `MemoryShardStore`: dense `Vec` → sparse `BTreeMap` (independent, small)

`MemoryShardStore` stores shards in a `Vec` indexed by shard index, and `put_shard` back-fills empty
shards from the last populated index up to the inserted one. Inserting a high-index shard — e.g. a wallet
synced from a **recent birthday**, whose first populated shard index is large — allocates a long run of
empty `LocatedTree`s: `O(max_index)` memory and `O(gap)` work, and it grows as the tree does.

A `BTreeMap<u64, LocatedPrunableTree<H>>` is sparse and (we believe) contract-compliant —
`ShardStore::get_shard_roots` is documented as "the addresses corresponding to the roots of subtrees
stored in this store", with no contiguity requirement. Direct heap win for `zcash_client_memory`/WebZjs,
and it also improves `CachingShardStore` (whose cache *is* a `MemoryShardStore`).

We'd add a test exercising **non-contiguous shard roots** to confirm nothing in the cap/frontier logic
assumes density. (Our production store uses exactly this `BTreeMap` representation, so we're fairly
confident — but it deserves an upstream test.)

## 2. `CachingShardStore`: dirty-delta + incremental flush

Today `flush()` writes the whole cached tree back. We propose:

[...]

PR #174, opened 2026-06-22
shardtree: store MemoryShardStore shards sparsely in a BTreeMap

LukasKorba:
Implements the first, independent part of #173: switch `MemoryShardStore`'s shard storage from a dense `Vec` to a sparse `BTreeMap` keyed by shard index.

### Why
`put_shard` back-fills an empty shard for every index below the highest inserted one, so inserting a high-index shard — e.g. a wallet synced from a recent birthday, whose first populated shard index is large — allocates a long run of empty shards (`O(max_index)` memory + `O(gap)` work). A `BTreeMap<u64, _>` stores only the shards actually inserted. Direct heap win for `zcash_client_memory` / WebZjs, and it also benefits `CachingShardStore` (whose cache is a `MemoryShardStore`).

### Behaviour preservation
Behaviour through the `ShardStore` interface is unchanged:

- Every `ShardTree` `get_shard` call site already tolerates a missing shard — `unwrap_or_else(|| empty)` (×3), `and_then`, `if let Some`, or `ok_or_else(|| …) → TreeIncomplete`. No site `.expect()`s a shard to be present.
- An empty (`Node::Nil`) shard's `root_hash` already returns the same `TreeIncomplete(vec![addr])` that a now-`None` shard produces via `ok_or_else`, so root/witness computations over a gap are identical.
- The only observable difference at the raw store API: a never-inserted index reads back from `get_shard` as `None` rather than a materialized empty shard (matching the SQLite-backed store), and `get_shard_roots` returns only populated indices.

### Tests
- New `non_contiguous_shards_are_sparse_and_root_correctly`: builds a tree with shards 2 & 3 present and 0 & 1 absent (a gap), then checks sparse `get_shard_roots`, `None` for the gap indices, correct `last_shard`, that the cap/frontier folds the two non-contiguous present shards into the `(3, 1)` root, and that a root spanning the gap reports `TreeIncomplete`.
- Full `shardtree` suite green: **86 passed** (the new test + 85 unchanged); `cargo fmt` / `clippy` clean.

This is PR 1 of the #173 overlay-store work; the `CachingShardStore` dirty-delta + incremental-flush piece (PR 2) follows once the seed-policy question in #173 is settled.

### DIFF
--- a/shardtree/src/store/memory.rs
+++ b/shardtree/src/store/memory.rs
@@ -1,19 +1,26 @@
 //! Implementation of an in-memory shard store with no persistence.
 
 use std::collections::BTreeMap;
-use std::convert::{Infallible, TryFrom};
+use std::convert::Infallible;
 
 use incrementalmerkletree::Address;
 
 use super::{Checkpoint, ShardStore};
-use crate::{LocatedPrunableTree, LocatedTree, PrunableTree, Tree};
+use crate::{LocatedPrunableTree, PrunableTree};
 
 /// An implementation of [`ShardStore`] that stores all state in memory.
 ///
 /// State is not persisted anywhere, and will be lost when the struct is dropped.
+///
+/// Shards are stored sparsely in a [`BTreeMap`] keyed by shard index: only shards explicitly
+/// inserted via [`ShardStore::put_shard`] are retained, and an index with no inserted shard reads
+/// back as `None` (never as an empty shard). This matches the behaviour of persistent backends such
+/// as the SQLite-backed store, and avoids materializing the run of empty shards below a high
+/// starting index — e.g. for a wallet synced from a recent birthday, whose first populated shard
+/// index is large.
 #[derive(Debug)]
 pub struct MemoryShardStore<H, C: Ord> {
-    shards: Vec<LocatedPrunableTree<H>>,
+    shards: BTreeMap<u64, LocatedPrunableTree<H>>,
     checkpoints: BTreeMap<C, Checkpoint>,
     cap: PrunableTree<H>,
 }
@@ -22,7 +29,7 @@ impl<H, C: Ord> MemoryShardStore<H, C> {
     /// Constructs a new empty `MemoryShardStore`.
     pub fn empty() -> Self {
         Self {
-            shards: vec![],
+            shards: BTreeMap::new(),
             checkpoints: BTreeMap::new(),
             cap: PrunableTree::empty(),
         }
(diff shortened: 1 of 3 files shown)


dannywillems (approved):
utACK 0099c4b1729968083953ab988be2385c0c242510

shardtree/src/store/memory.rs
```diff
@@ -1,19 +1,26 @@
...
-use crate::{LocatedPrunableTree, LocatedTree, PrunableTree, Tree};
+use crate::{LocatedPrunableTree, PrunableTree};
 
 /// An implementation of [`ShardStore`] that stores all state in memory.
 ///
 /// State is not persisted anywhere, and will be lost when the struct is dropped.
+///
+/// Shards are stored sparsely in a [`BTreeMap`] keyed by shard index: only shards explicitly
+/// inserted via [`ShardStore::put_shard`] are retained, and an index with no inserted shard reads
+/// back as `None` (never as an empty shard). This matches the behaviour of persistent backends such
```
dannywillems, line 17:
```
This matches the behaviour of persistent backends such
/// as the SQLite-backed store, and avoids materializing the run of empty shards below a high
/// starting index — e.g. for a wallet synced from a recent birthday, whose first populated shard
/// index is large.
```

This comment should not appear in this library, as `shardtree` should not be aware/care of any user implementation details.

shardtree/src/store/memory.rs
```diff
@@ -38,39 +45,25 @@ impl<H: Clone, C: Clone + Ord> ShardStore for MemoryShardStore<H, C> {
...
-
-        let shard_idx =
-            usize::try_from(subtree_addr.index()).expect("SHARD_HEIGHT > 64 is unsupported");
-        self.shards[shard_idx] = subtree;
+        self.shards.insert(subtree.root_addr.index(), subtree);
         Ok(())
     }
 
     fn get_shard_roots(&self) -> Result<Vec<Address>, Self::Error> {
-        Ok(self.shards.iter().map(|s| s.root_addr).collect())
```
dannywillems, line 68: Mmh, I already merged, but this changed the semantic of `get_shard_roots`. Checking...

shardtree/src/store/memory.rs
```diff
@@ -1,19 +1,26 @@
...
+/// Shards are stored sparsely in a [`BTreeMap`] keyed by shard index: only shards explicitly
+/// inserted via [`ShardStore::put_shard`] are retained, and an index with no inserted shard reads
+/// back as `None` (never as an empty shard). This matches the behaviour of persistent backends such
+/// as the SQLite-backed store, and avoids materializing the run of empty shards below a high
+/// starting index — e.g. for a wallet synced from a recent birthday, whose first populated shard
+/// index is large.
 #[derive(Debug)]
 pub struct MemoryShardStore<H, C: Ord> {
-    shards: Vec<LocatedPrunableTree<H>>,
+    shards: BTreeMap<u64, LocatedPrunableTree<H>>,
```
nuttycom, line 23: nit: this is more space than we need for a key, because we'd need a 128-bit position type. I would have made the key a u32.

nuttycom: I guess that the address index uses u64 though so this is fine. It's probably the preexisting structure of `Address` that's greedy.

shardtree/src/store/memory.rs
nuttycom: Yes, this is a breaking change and actually quite a bad one, we shouldn't make this change.

dannywillems: Done in https://github.com/zcash/incrementalmerkletree/pull/179

PR #175, opened 2026-06-22
shardtree: flush only the delta from CachingShardStore

LukasKorba:
Implements the second part of #173 (approach ACK'd): make `CachingShardStore` flush only the delta, and add an incremental flush.

### Why
`flush` currently rewrites the *entire* cached tree to the backend — every shard, the cap, every checkpoint — on every call, regardless of what changed (`O(tree)` work per flush). For a consumer that flushes repeatedly during a long sync, or persists to a slow/async backend (e.g. `zcash_client_memory` / WebZjs → IndexedDB), that rewrite is the dominant cost.

### What
- `CachingShardStore` now tracks which shards and checkpoints (and the cap) were modified since it was loaded.
- `flush` writes only that delta. **The resulting backend state is unchanged** — transparent to existing callers.
- New `flush_delta(&mut self)`: writes the delta and leaves the store usable, for callers that flush incrementally (e.g. once per scanned chunk) instead of once at the end. `flush(self)` is now `flush_delta()` + returning the backend.

### Correctness
Deferred truncations/removals are applied first, then modified shards/checkpoints that still exist in the cache are written; a modification that was later truncated away is skipped (the deferred truncation already removed it from the backend). This preserves the existing flush ordering.

- The existing `check_equal` tests assert that a flushed `CachingShardStore` backend is **byte-identical** to a directly-mutated `MemoryShardStore` across the full `check_operations` suite (appends, checkpoints, witnesses, rewinds/truncations) — so they gate the delta's result-preservation.
- New `flush_delta_composes_across_incremental_flushes`: flushes incrementally partway through, asserts the dirty state is reset, mutates further, then verifies the final backend still matches the reference — proving incremental flushes compose.
- Full `shardtree` suite green; `cargo fmt` / `clippy` clean.

Independent of #174 (that PR touches `memory.rs`; this one only `caching.rs`) — they can land in either order. Together they form the overlay-store enhancement in #173; the explicit-preload seed policy follows as a separate PR.

### DIFF
--- a/shardtree/src/store/caching.rs
+++ b/shardtree/src/store/caching.rs
@@ -1,5 +1,6 @@
 //! Implementation of an in-memory shard store with persistence.
 
+use std::collections::{BTreeMap, BTreeSet};
 use std::convert::Infallible;
 
 use incrementalmerkletree::Address;
@@ -28,6 +29,12 @@ where
     backend: S,
     cache: MemoryShardStore<S::H, S::CheckpointId>,
     deferred_actions: Vec<Action<S::CheckpointId>>,
+    /// Root addresses of shards written since load / the last flush, keyed by shard index.
+    dirty_shards: BTreeMap<u64, Address>,
+    /// Whether the cap has been written since load / the last flush.
+    cap_dirty: bool,
+    /// Identifiers of checkpoints added or updated since load / the last flush.
+    dirty_checkpoints: BTreeSet<S::CheckpointId>,
 }
 
 impl<S> CachingShardStore<S>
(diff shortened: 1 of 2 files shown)


shardtree/src/store/caching.rs
```diff
@@ -58,60 +65,69 @@ where
...
-                    self.backend.remove_checkpoint(checkpoint_id)
+                    self.backend.remove_checkpoint(&checkpoint_id)
                 }
                 Action::TruncateCheckpointsRetaining(checkpoint_id) => {
-                    self.backend.truncate_checkpoints_retaining(checkpoint_id)
+                    self.backend.truncate_checkpoints_retaining(&checkpoint_id)
                 }
             }?;
         }
-        self.deferred_actions.clear();
```
dannywillems, line 77: This disappeared. Is it expected?

dannywillems: Ah, I see, std::mem::take replaces it. It is its side-effect.

dannywillems (approved):
LGTM, but how do you plan to use this diff? At the moment, it only keeps a diff, and keep the same semantic.

dannywillems (changes requested):
Blocking the merge until https://github.com/zcash/incrementalmerkletree/pull/181/changes/f8cede353aed2b6fb2a0e238449caffb8b25cad2#r3473444582 is resolved.

LukasKorba:
superseded by [#181](https://github.com/zcash/incrementalmerkletree/pull/181)

PR #178, opened 2026-06-24
shardtree: add regression test for get_shard_roots consumers

dannywillems:
PR #174 made MemoryShardStore store shards sparsely in a BTreeMap, so get_shard_roots no longer reports back-filled empty placeholder shards below the first populated index. Its only consumer, marked_positions, must keep reporting every mark from populated shards and must not fabricate positions for the absent gap.

Add marked_positions_consumer_tolerates_sparse_shard_roots, which populates non-contiguous shards with marks above an absent gap and asserts both that get_shard_roots stays sparse and that marked_positions returns exactly the marks from the populated shards.


dannywillems:
Closing in favor of https://github.com/zcash/incrementalmerkletree/pull/180

PR #180, opened 2026-06-24
shardtree: generalize proptest strategies and add arb_shard_layout

dannywillems:
Add size-parametric proptest strategies so property tests are no longer locked to a single tree geometry, and a strategy for arbitrary shard layouts that replaces hand-rolled fixtures.

- arb_leaves_sized(arb_leaf, max_count): arb_leaves now delegates to it with the previous 2^6 bound.
- arb_shardtree_sized::<H, DEPTH, SHARD_HEIGHT>: const-generic over the tree dimensions; arb_shardtree delegates to the 6,3 instantiation.
- arb_shard_layout(arb_leaf, shard_height, max_index, max_shards): generates complete shards at distinct, possibly non-contiguous indices with random per-leaf retention, paired with the marked positions they contain as an oracle.

Rewrite check_shardtree_caching onto arb_shardtree_sized via a geometry-generic check_caching helper, and exercise it at a second 4,2 geometry. Add marked_positions_matches_shard_layout to cover arb_shard_layout against ShardTree::marked_positions.

Also add a regtest, as initially it was for https://github.com/zcash/incrementalmerkletree/pull/178.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -2035,6 +2035,24 @@ mod tests {
             }
             prop_assert_eq!(tree.marked_positions().unwrap(), layout.marked_positions);
         }
+
+        // Regression test: the dense MemoryShardStore back-fills empty
+        // placeholder shards, so get_shard_roots reports a contiguous
+        // 0..=max_populated_index range of roots.
+        #[test]
+        fn get_shard_roots_is_dense_up_to_max_index(
+            layout in arb_shard_layout(arb_char_str(), 2, 8, 4)
+        ) {
+            let mut tree = empty_tree::<String, 4, 2>();
+            for shard in &layout.shards {
+                tree.store.put_shard(shard.clone()).unwrap();
+            }
+            let max_index = layout.shards.iter().map(|s| s.root_addr.index()).max().unwrap();
+            let expected: Vec<Address> = (0..=max_index)
+                .map(|i| Address::from_parts(Level::from(2), i))
+                .collect();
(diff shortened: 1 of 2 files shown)


shardtree/src/lib.rs
```diff
@@ -1981,31 +1981,77 @@ mod tests {
...
+        fn check_shardtree_caching_small_shards(
+            (tree, _, marked_positions) in arb_shardtree_sized::<_, 4, 2>(arb_char_str())
+        ) {
+            check_caching(tree, marked_positions);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn marked_positions_matches_shard_layout(
```
nuttycom, line 2029: For proptests in particular, I think that we should specifically call out the properties under test in a comment, since these are intended to be universal properties that hold regardless of inputs. It's often somewhat difficult to identify what the actual contract that's established by the proptest is just by reading the code.

shardtree/src/lib.rs
```diff
@@ -1981,31 +1981,77 @@ mod tests {
...
+        ) {
+            check_caching(tree, marked_positions);
+        }
+
+        #[test]
+        fn check_shardtree_caching_small_shards(
+            (tree, _, marked_positions) in arb_shardtree_sized::<_, 4, 2>(arb_char_str())
+        ) {
+            check_caching(tree, marked_positions);
+        }
```
nuttycom, line 2024: The difference between this and the test above seems pretty trivial; why are both needed?

nuttycom (approved):
utACK with non-blocking nits.

PR #181, opened 2026-06-24
shardtree: add SparseCachingShardStore (self-contained sparse overlay)

LukasKorba:
Implements the explicit-preload seed policy from #173 (its third piece). This supersedes the earlier version that stacked on #175 — per review, #175 is dropped and `SparseCachingShardStore` is now **fully self-contained**, leaving `CachingShardStore` unchanged.

## What

`CachingShardStore::load` copies the entire backend into the cache (`O(tree)` memory + work — costly for a store synced from a recent birthday, or an async/wasm backend where each read is expensive). `SparseCachingShardStore::with_preloaded(backend, shards)` copies only an explicitly provided working set of shards (plus the cap and checkpoints), and on flush writes back only the delta accumulated since preload — so the in-memory overlay is `O(working set)` rather than `O(tree)` per checkpoint (the win for `zcash_client_memory`/WebZjs, and our wallet sync).

It maintains its **own** cache, dirty-tracking, deferred truncations and delta flush (`flush` / non-consuming `flush_delta`); it does **not** wrap `CachingShardStore`. `CachingShardStore` is left entirely unchanged by this PR — eager, `Infallible`, whole-cache flush exactly as on `main`. The only edit in its region of the file is a `use std::collections::{BTreeMap, BTreeSet}` import the sparse store consumes.

## Strict read-miss

The only behavioural difference from the eager store is a strict read-miss policy: `get_shard` for a shard that exists in the backend but was not preloaded returns `SparseStoreError::NotPreloaded(idx)` rather than `None`. Returning `None` there would silently corrupt tree operations — the append path builds on an assumed-empty base and overwrites the real shard on flush — so the miss fails loudly. Because of this the error type is not `Infallible`. The caller's contract is to preload every shard its operations will touch, including the frontier (last) shard on which appends build.

## Tests

- `sparse_store_matches_eager_on_fresh_appends` — a sparse store seeded from an empty backend matches the eager reference (`CombinedTree`) across appends + a checkpoint, and its delta flush reproduces the same backend state.
- `sparse_store_strict_read_miss_policy` — preloaded shard → hit; backend-resident-but-unpreloaded → `NotPreloaded`; absent → `None`; `get_shard_roots` → union of cached + backend shards.
- `sparse_flush_delta_composes_across_incremental_flushes` — an intermediate non-consuming `flush_delta`, then more mutations + a final flush, reproduces a directly-mutated reference, proving incremental flushes compose and the dirty state resets correctly.

`CachingShardStore`'s existing suite (append / witnesses / rewind / checkpoint) continues to pass unchanged, gating the restore. Full `shardtree` suite green (**88 passed** + 4 doc-tests); `cargo fmt` / `clippy` clean.

### DIFF
--- a/shardtree/src/store/caching.rs
+++ b/shardtree/src/store/caching.rs
@@ -1,6 +1,6 @@
 //! Implementation of an in-memory shard store with persistence.
 
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::convert::Infallible;
 
 use incrementalmerkletree::Address;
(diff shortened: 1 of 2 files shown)


shardtree/src/store/caching.rs
```diff
@@ -256,14 +263,312 @@ where
...
+    S::CheckpointId: Clone + Ord,
+{
+    inner: CachingShardStore<S>,
+    /// Root addresses of the shards present in the backend at preload time, keyed by shard index.
+    /// Distinguishes "exists in the backend but not preloaded" (→ [`SparseStoreError::NotPreloaded`])
+    /// from "absent everywhere" (→ `None`).
+    backend_shard_roots: BTreeMap<u64, Address>,
+}
+
+impl<S> SparseCachingShardStore<S>
```
dannywillems, line 346: Nit: I would move this in its own module unit, `sparse_caching` maybe?

EDIT: maybe it is too picky. Let's ignore.

shardtree/src/store/caching.rs
```diff
@@ -256,14 +263,312 @@ where
...
+/// frontier (last) shard**, on which appends build.
+pub struct SparseCachingShardStore<S>
+where
+    S: ShardStore,
+    S::H: Clone,
+    S::CheckpointId: Clone + Ord,
+{
+    inner: CachingShardStore<S>,
+    /// Root addresses of the shards present in the backend at preload time, keyed by shard index.
+    /// Distinguishes "exists in the backend but not preloaded" (→ [`SparseStoreError::NotPreloaded`])
```
dannywillems: There are multiple instances of UTF-8 characters. Can you remove them/replace them with standard ASCII, please?

shardtree/src/store/caching.rs
```diff
@@ -256,14 +263,312 @@ where
...
+///
+/// The caller is responsible for preloading every shard its operations will touch — **including the
+/// frontier (last) shard**, on which appends build.
+pub struct SparseCachingShardStore<S>
+where
+    S: ShardStore,
+    S::H: Clone,
+    S::CheckpointId: Clone + Ord,
+{
+    inner: CachingShardStore<S>,
```
dannywillems: Can you explain why you added the additional fields directly in `CachingShardStore` if it is used only here, and in addition to that, if the trait implementation below is essentially calling the methods of the inner structure?
Instead of changing the already used structure above, maybe we could keep it the same, and move all the fields here.

shardtree/src/store/caching.rs
```diff
@@ -256,14 +263,312 @@ where
...
+        shard_root: Address,
+    ) -> Result<Option<LocatedPrunableTree<Self::H>>, Self::Error> {
+        // A cache hit (preloaded, or created since preload) is returned directly.
+        if let Some(shard) = self.inner.get_shard(shard_root).map_err(|e| match e {})? {
+            return Ok(Some(shard));
+        }
+        // Not cached: if the backend holds this shard, the caller failed to preload it. Returning
+        // `None` here would silently corrupt tree operations, so fail loudly instead.
+        if self.backend_shard_roots.contains_key(&shard_root.index()) {
+            return Err(SparseStoreError::NotPreloaded(shard_root.index()));
```
dannywillems, line 491: Why does it fail if it is in the backend?

shardtree/src/store/caching.rs
LukasKorba: Agreed — leaving it inline per your edit. Easy to split into a sparse_caching submodule later if the file grows.

shardtree/src/store/caching.rs
LukasKorba: The dirty-tracking fields + flush_delta are a general enhancement to CachingShardStore, not sparse-only — flush_delta (incremental, non-consuming delta flush) is independently useful for the full store (preload-all → mutate → flush only the delta, vs flush rewriting everything). They live here because the dirty-marking must happen inside the write methods (put_shard/put_cap/add_checkpoint/truncate), where the cache actually mutates; SparseCachingShardStore then reuses that by composition (it delegates writes to inner). Moving the tracking into SparseCachingShardStore would mean re-implementing those four writes to  intercept-then-delegate, and flush_delta would need CachingShardStore's private cache/backend to write back — leaking internals. So the wrapper stays thin (just the strict read-miss policy + backend_shard_roots) precisely because the reusable delta machinery is on CachingShardStore. Can add a CHANGELOG line making the general flush_delta explicit.

shardtree/src/store/caching.rs
LukasKorba: Strict-preload contract: the store only holds shards the caller preloaded, and the caller must preload every shard their ops touch. A get_shard miss for a shard that exists in the backend means a preload was forgotten — returning None would make the tree treat a non-empty shard as absent → silently wrong roots / lost notes on the next mutation. So we fail loudly with NotPreloaded(index) to surface it immediately. (A miss for a shard absent from the backend too is genuinely None.)

shardtree/src/store/caching.rs
LukasKorba: Done in the latest push (9c9f374) -- replaced the em-dashes (U+2014) and arrows (U+2192) in the doc comments with ASCII (-- / ->). rg '[^\x00-\x7F]' shardtree/src/store/caching.rs returns nothing now.

shardtree/src/store/caching.rs
LukasKorba: Because returning None there would be silent corruption, not a valid "not found". The sparse store only caches the working set you explicitly preload; a shard that's in the backend but wasn't preloaded is a caller bug (under-preload), not an absent shard. If get_shard returned None for it, the append path would treat that subtree as empty, build on top of it, and overwrite the real shard on the next flush — losing data with no signal. The eager CachingShardStore never has to distinguish the two cases because it loads everything, so None always means truly-absent there; the sparse store can't assume that, so it surfaces the under-preload loudly as NotPreloaded instead. The contract is just: preload every shard your ops will touch (including the frontier shard appends build on), which is the O(working-set) you're paying for anyway.

nuttycom (approved):
utACK fb2e07209036abd493000172c55a578186e58fd2

PR #182, opened 2026-06-26
Split ShardStore into a truncable and one not truncable

dannywillems:


nuttycom:
Closing as not needed; #181 was updated to support truncation.

PR #184, opened 2026-06-30
shardtree: explicit retention of checkpoints (durable anchors)

### LINKED ISSUE #183: shardtree: explicit retention of checkpoints (durable anchors queryable past max_checkpoints)
## Summary

Allow a checkpoint to be explicitly **retained** so it is exempt from automatic pruning of excess checkpoints. A retained checkpoint's root — and witnesses for `Marked` leaves at or before it — remain computable even after it has aged more than `max_checkpoints` behind the tip of the tree. Retention is releasable.

## Motivation

A caller (e.g. a wallet) sometimes needs a durable "anchor": the ability to compute the tree root and witnesses as of a specific historical checkpoint that is older than the rolling `max_checkpoints` window. Previously such a checkpoint's record was unconditionally removed once `checkpoint_count > max_checkpoints`, so its id no longer resolved.

## Design (as implemented)

Retention is an explicit operation, **decoupled from leaf marking**:

- **`ShardTree::ensure_retained(checkpoint_id)`** records an id in a new explicit retention set on the store; **`ShardTree::remove_retained_checkpoint(&id)`** releases it. (The id may be recorded even if no such checkpoint currently exists.)
- **`ShardStore`** gains `add_retained_checkpoint` / `remove_retained_checkpoint` / `retained_checkpoints` (the id-keyed retention set). Implemented for `MemoryShardStore` and the caching store.
- **`prune_excess_checkpoints`** excludes retained checkpoints from the `max_checkpoints` budget and never auto-removes them. It loads the retention set once (no per-checkpoint store lookup).

### Why no `Reference` retention is needed

Computing a witness for a marked leaf as of checkpoint `R` truncates the tree at `R+1`. The only subtree that truncation cuts through is the one straddling position `R` — and that subtree contains the anchor's own leaf, whose **`CHECKPOINT`** retention (kept because the checkpoint is retained) propagates through merges and prevents that subtree from collapsing. Every other sibling subtree is either wholly ≤ `R` (its collapsed aggregate hash is valid) or wholly > `R` (forced empty). So `CHECKPOINT` + record retention is sufficient; `REFERENCE`/leaf-marking is not required for this feature. (Verified empirically: a plain retained checkpoint yields the identical witness.)

## Downstream impact (librustzcash)

- **SQLite (`zcash_client_sqlite`):** a small retained-checkpoint-ids table + implementations of the three new `ShardStore` retention-set methods in `commitment_tree.rs`. Decoupled from the checkpoint record, so no change to the existing checkpoint tables. Only needed going forward.
- **In-memory (`zcash_client_memory`):** reuses `MemoryShardStore`, which carries the retention set for free.
- A wallet that wants a durable anchor calls `ensure_retained(checkpoint_id)` on the relevant checkpoint.

## Status

Implemented on branch `retained_reference_checkpoints` (single commit). Tests (all green): `anchor_checkpoint_survives_pruning`, `witness_marked_leaf_as_of_anchor_checkpoint` (the motivating scenario — a marked leaf witnessed as of an aged-out anchor), `released_anchor_checkpoint_is_pruned`.

## Related

The investigation behind this feature surfaced a separate concern — the wallet's automatic frontier `Marking::Reference` retention accumulates unbounded un-prunable nodes during forward sync: zcash/librustzcash#2503.

---
*Filed and updated with the assistance of Claude Code.*

nuttycom:
## Summary

Adds the ability to explicitly **retain** a checkpoint so that it is exempt from automatic pruning of excess checkpoints. A retained checkpoint's root — and witnesses for `Marked` leaves at or before it — remain computable even after it has aged more than `max_checkpoints` behind the tip of the tree. Retention is releasable.

This supersedes the earlier draft #95 ("Add the ability to avoid pruning specific checkpoints"): same explicit-retention idea, rebased onto current `main` and simplified (no `Reference`/marking changes, no public `incrementalmerkletree` API change).

## API

- `ShardTree::ensure_retained(checkpoint_id)` — mark a checkpoint for retention (the id may be recorded even if no such checkpoint currently exists).
- `ShardTree::remove_retained_checkpoint(&checkpoint_id)` — release it.
- `ShardStore`: `add_retained_checkpoint` / `remove_retained_checkpoint` / `retained_checkpoints` (the id-keyed retention set; implemented for `MemoryShardStore` and the caching store).

`prune_excess_checkpoints` excludes retained checkpoints from the `max_checkpoints` budget and never auto-removes them; it loads the retention set once rather than querying per checkpoint.

## Why no `Reference` retention is needed

Computing a witness for a marked leaf as of checkpoint `R` truncates the tree at `R + 1`. The only subtree truncation cuts through is the one straddling position `R`, which contains the anchor's own leaf — and its `CHECKPOINT` retention (kept because the checkpoint is retained) propagates through merges and prevents that subtree from collapsing. Every other sibling subtree is either wholly ≤ `R` (its aggregate hash is valid) or wholly > `R` (forced empty). So record retention plus the inherent `CHECKPOINT` protection is sufficient; no `Reference`/leaf marking is required, and the public `incrementalmerkletree` `Marking`/`Retention` API is unchanged.

## Tests

- `anchor_checkpoint_survives_pruning` — a retained checkpoint survives pruning that would remove an ordinary checkpoint in the same position.
- `witness_marked_leaf_as_of_anchor_checkpoint` — the motivating scenario: a `Marked` leaf is witnessed (and the root computed) as of a retained checkpoint that has aged more than `max_checkpoints` behind the tip.
- `released_anchor_checkpoint_is_pruned` — after `remove_retained_checkpoint`, the checkpoint is pruned normally.

## Related

The investigation behind this surfaced a separate concern in the wallet's use of frontier `Marking::Reference` retention (unbounded accumulation during forward sync): zcash/librustzcash#2503.

Closes #183.

### DIFF
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -13,6 +13,20 @@ and this project adheres to Rust's notion of
 - `shardtree::LocatedTree::pretty_print_bottom_top_with`
 - `shardtree::LocatedPrunableTree::pretty_print_indented`
 - `shardtree::LocatedPrunableTree::pretty_print_bottom_top`
+- `shardtree::ShardTree::ensure_retained`
+- `shardtree::ShardTree::remove_retained_checkpoint`
+- `shardtree::store::ShardStore::add_retained_checkpoint`
+- `shardtree::store::ShardStore::remove_retained_checkpoint`
+- `shardtree::store::ShardStore::retained_checkpoints`
+
+### Changed
+- A checkpoint that has been explicitly retained via
+  `ShardTree::ensure_retained` is treated as a durable "anchor":
+  `ShardTree::prune_excess_checkpoints` excludes retained checkpoints from the
+  `max_checkpoints` budget and never removes them, so their root and the
+  witnesses for marked leaves at or before them remain computable even after
+  they have aged more than `max_checkpoints` behind the tip of the tree.
+  Retention is released with `ShardTree::remove_retained_checkpoint`.
 
 ## [0.6.2] - 2026-02-20
 
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.7.0] - PLANNED
 
 ### Added
 - `impl Display for shardtree::RetentionFlags`
@@ -15,9 +15,6 @@ and this project adheres to Rust's notion of
 - `shardtree::LocatedPrunableTree::pretty_print_bottom_top`
 - `shardtree::ShardTree::ensure_retained`
 - `shardtree::ShardTree::remove_retained_checkpoint`
-- `shardtree::store::ShardStore::add_retained_checkpoint`
-- `shardtree::store::ShardStore::remove_retained_checkpoint`
-- `shardtree::store::ShardStore::retained_checkpoints`
 
 ### Changed
 - A checkpoint that has been explicitly retained via
@@ -27,6 +24,11 @@ and this project adheres to Rust's notion of
   witnesses for marked leaves at or before them remain computable even after
   they have aged more than `max_checkpoints` behind the tip of the tree.
   Retention is released with `ShardTree::remove_retained_checkpoint`.
+- The `shardtree::store::ShardStore` trait has been modified. It has added methods:
+  - `ShardStore::add_retained_checkpoint`
+  - `ShardStore::remove_retained_checkpoint`
+  - `ShardStore::retained_checkpoints`
+
 
 ## [0.6.2] - 2026-02-20
 
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -577,6 +577,32 @@ impl<
         Ok(true)
     }
 
+    /// Marks the checkpoint having the given identifier for retention, exempting it from automatic
+    /// pruning of excess checkpoints.
+    ///
+    /// A retained checkpoint is excluded from the `max_checkpoints` budget and is never removed by
+    /// [`Self::prune_excess_checkpoints`]; it persists until it is released via
+    /// [`Self::remove_retained_checkpoint`]. As a result, its root, and witnesses for marked leaves
+    /// at or before it, remain computable even after it has aged more than `max_checkpoints` behind
+    /// the tip of the tree. The identifier may be recorded even if no checkpoint with that
+    /// identifier currently exists.
+    pub fn ensure_retained(&mut self, checkpoint_id: C) -> Result<(), ShardTreeError<S::Error>> {
+        self.store
+            .add_retained_checkpoint(checkpoint_id)
+            .map_err(ShardTreeError::Storage)
+    }
+
+    /// Releases a checkpoint previously retained via [`Self::ensure_retained`], allowing it to be
+    /// pruned normally. Has no effect if the checkpoint was not retained.
+    pub fn remove_retained_checkpoint(
+        &mut self,
+        checkpoint_id: &C,
+    ) -> Result<(), ShardTreeError<S::Error>> {
+        self.store
+            .remove_retained_checkpoint(checkpoint_id)
+            .map_err(ShardTreeError::Storage)
+    }
+
     /// Removes the oldest checkpoints until at most `max_checkpoints` remain,
     /// clearing the `CHECKPOINT` and `MARKED` retention flags they were keeping
     /// alive so the affected leaves become prunable.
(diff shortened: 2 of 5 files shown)


shardtree/CHANGELOG.md
```diff
@@ -13,6 +13,20 @@ and this project adheres to Rust's notion of
 - `shardtree::LocatedTree::pretty_print_bottom_top_with`
 - `shardtree::LocatedPrunableTree::pretty_print_indented`
 - `shardtree::LocatedPrunableTree::pretty_print_bottom_top`
+- `shardtree::ShardTree::ensure_retained`
+- `shardtree::ShardTree::remove_retained_checkpoint`
+- `shardtree::store::ShardStore::add_retained_checkpoint`
+- `shardtree::store::ShardStore::remove_retained_checkpoint`
+- `shardtree::store::ShardStore::retained_checkpoints`
```
nuttycom: Since these are a trait change, they need to go in the `Changed` section.
```suggestion
```

shardtree/CHANGELOG.md
```diff
@@ -13,6 +13,20 @@ and this project adheres to Rust's notion of
...
+
+### Changed
+- A checkpoint that has been explicitly retained via
+  `ShardTree::ensure_retained` is treated as a durable "anchor":
+  `ShardTree::prune_excess_checkpoints` excludes retained checkpoints from the
+  `max_checkpoints` budget and never removes them, so their root and the
+  witnesses for marked leaves at or before them remain computable even after
+  they have aged more than `max_checkpoints` behind the tip of the tree.
+  Retention is released with `ShardTree::remove_retained_checkpoint`.
```
nuttycom, line 26:
```suggestion
  Retention is released with `ShardTree::remove_retained_checkpoint`.
- The `shardtree::store::ShardStore` trait has been modified. It has added methods:
  - `ShardStore::add_retained_checkpoint`
  - `ShardStore::remove_retained_checkpoint`
  - `ShardStore::retained_checkpoints`

```

shardtree/CHANGELOG.md
nuttycom:
```suggestion
## [0.7.0] - PLANNED
```

shardtree/src/lib.rs
```diff
@@ -590,15 +616,35 @@ impl<
...
             .map_err(ShardTreeError::Storage)?;
+
+        // Checkpoints in the explicit retention set are pinned: they are excluded from the
+        // `max_checkpoints` budget and are never removed by automatic pruning. The budget therefore
+        // applies only to the non-retained checkpoints.
+        let retained = self
+            .store
+            .retained_checkpoints()
+            .map_err(ShardTreeError::Storage)?;
+        let mut retained_existing = 0usize;
```
dannywillems, line 627: Review comment: I was initially thinking that this was equivalent to caling `len` on the `BTreeSet`, but the invariant that the BTreeSet is a subset of the checkpoints might be valid at runtime, but not at compile time; there might be some old values in the BTreeSet. Therefore it is safer to have this runtime logic.

dannywillems: Note: complexity cost is negligible. LGTM.

shardtree/src/store/memory.rs
```diff
@@ -166,6 +168,20 @@ impl<H: Clone, C: Clone + Ord> ShardStore for MemoryShardStore<H, C> {
         Ok(())
     }
 
+    fn add_retained_checkpoint(&mut self, checkpoint_id: C) -> Result<(), Self::Error> {
+        self.retained_checkpoints.insert(checkpoint_id);
+        Ok(())
+    }
+
+    fn remove_retained_checkpoint(&mut self, checkpoint_id: &C) -> Result<(), Self::Error> {
+        self.retained_checkpoints.remove(checkpoint_id);
```
dannywillems, line 177: Review notes: `remove` returns a boolean, informing if a value was removed or not. I notice that we don't enforce the binding (even if ignored) of the returned values, when a method is defined with `#[must_use]`. I've asked Claude to fix this, in case we are interested in enforcing it. IMO, it is a good practice, in particular in this AI area where we will need to make more informed reviews.

dannywillems (approved):
utACK 9ac65ec03c970d2fd54daf0ef4e75dc2b13c8574

shardtree/src/store/memory.rs
sellout: I agree it should be enforced. I really think it should go the other way – `must_use` by default with some attribute to make use optional.

shardtree/src/store/caching.rs
```diff
@@ -111,6 +116,14 @@ where
             self.backend.add_checkpoint(checkpoint_id, checkpoint)?;
         }
 
+        for checkpoint_id in self
+            .cache
+            .retained_checkpoints()
+            .expect("error type is Infallible")
```
sellout, line 122: If the error is `Infallable`, shouldn’t there be something like `impl<T> From<Result<T, !>> for T` that we can take advantage of here, rather than using `expect`?

That would also increase safety, as with the current code, if the error type changes, this still compiles just fine, but using `From` wouldn’t.

sellout (commented):
Post-hoc utACK.

One question/suggestion, and one comment in support of @dannywillems’ suggestion.

shardtree/src/store/memory.rs
nuttycom: Yeah I'm fine with `must_use` being more strictly enforced.

PR #187, opened 2026-07-02
Must use lint

### LINKED ISSUE #186: Configure clippy with must_use_candidate = deny.
> Review notes: `remove` returns a boolean, informing if a value was removed or not. I notice that we don't enforce the binding (even if ignored) of the returned values, when a method is defined with `#[must_use]`. I've asked Claude to fix this, in case we are interested in enforcing it. IMO, it is a good practice, in particular in this AI area where we will need to make more informed reviews. 

 _Originally posted by @dannywillems in [#184](https://github.com/zcash/incrementalmerkletree/pull/184/changes#r3512073725)_

dannywillems:


nuttycom:
I feel like all of the `let _ = foo.insert(...)` do not add value; these are calls that are very frequently made just for their side effects, where the caller is indifferent to the result. I think that maybe this PR goes farther than it needs to?

dannywillems:
> I feel like all of the `let _ = foo.insert(...)` do not add value; these are calls that are very frequently made just for their side effects, where the caller is indifferent to the result. I think that maybe this PR goes farther than it needs to?

It is the expected behavior, and changes that this patch aims to introduce.
[BTreeSet::insert](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.insert) returns a boolean, describing if a value has been inserted or not.

> Adds a value to the set.
> Returns whether the value was newly inserted. That is:
> - If the set did not previously contain an equal value, true is returned.
> - If the set already contained an equal value, false is returned, and the entry is not updated.

The output is actually unused in our case, and writing in the code `let _ = [...]` gives the reader the information that the methods have some hypothesis, like, in the case of `BTreeSet::insert`: "the value MIGHT be already present in the set", and an expected outcome could be: "and the behavior must be the same, either with an existing value or not".

A reader (human or, nowadays, AI agents) who has this context while reading code can therefore reason about the semantic of the routines while looking at the code. An ignored let-binding to a non-unit value hides some semantics.

However, I understand that the code is a bit (relatively) heavy to read. Personally, I have this habit (of not ignoring let-binding to non-unit values) from OCaml as, by default, the compiler adds a warning. For instance, you can test it with the following code:

```ocaml
let f x = x * x;;

let () =
  f 5;
  print_endline "Hello, World"
```

on [Try OCaml](https://try.ocamlpro.com/)

dannywillems:
Closing as, after discussion, this patch adds too much noise, and I don't have strong opinion.

PR #189, opened 2026-07-06
shardtree: seed CachingShardStore's retention set at load

### LINKED ISSUE #188: `CachingShardStore::load` drops the backend's retained-checkpoint set
#184 added retained checkpoints ("anchors") to `ShardStore`: `ShardTree::prune_excess_checkpoints` consults `retained_checkpoints()` to decide which checkpoints it may remove, so a retained checkpoint is never pruned.

`CachingShardStore::load` copies shards, the cap and checkpoints into its in-memory cache, but **not the backend's retention set**. As a result, `retained_checkpoints()` on a freshly-loaded overlay reports `{}` regardless of backend state.

**Consequences:**
- `prune_excess_checkpoints` performed through a freshly-loaded overlay treats every checkpoint as prunable, so a durable anchor's checkpoint can be removed; the deferred `RemoveCheckpoint` action then replays onto the backend at flush.
- `flush` writes the *cache's* retention set back to the backend, so after a load → prune → flush cycle the backend can hold a retention marker pointing at a checkpoint that no longer exists (tolerated by the trait contract, but the anchor's durability guarantee is silently lost).
- Even without pruning, `retained_checkpoints()` through the overlay simply answers incorrectly.

**Repro:** load a backend that has a retained checkpoint; `retained_checkpoints()` on the overlay returns the empty set. (Regression test included in the fix PR.)

**Fix:** seed the cache's retention set from the backend in `load`, mirroring how shards/cap/checkpoints are copied — one loop. PR follows.

Found while rebasing #181, whose `SparseCachingShardStore` needed the same retained-set handling for its preload path.

LukasKorba:
Fixes #188.

`CachingShardStore::load` copied shards, the cap and checkpoints into its in-memory cache, but not the backend's retained-checkpoint set, so `retained_checkpoints()` on a freshly-loaded overlay reported an empty set regardless of backend state — and pruning performed through the overlay could therefore remove a durable anchor's checkpoint (the deferred removal then replays onto the backend at flush).

The fix seeds the cache's retention set from the backend at load, mirroring how shards, the cap and checkpoints are copied. A regression test covers both the load itself and the load → flush round trip.

No changelog entry: this repairs behavior introduced by #184, which is unreleased.

Found while rebasing #181 (`SparseCachingShardStore` handles its retention set the same way in `with_preloaded`).

### DIFF
--- a/shardtree/src/store/caching.rs
+++ b/shardtree/src/store/caching.rs
@@ -55,6 +55,13 @@ where
                 .expect("error type is Infallible");
             Ok(())
         })?;
+        // Seed the retention set so that pruning performed through the overlay observes the
+        // backend's durable anchors (a retained checkpoint must never be pruned).
+        for checkpoint_id in backend.retained_checkpoints()? {
+            cache
+                .add_retained_checkpoint(checkpoint_id)
+                .expect("error type is Infallible");
+        }
 
         Ok(Self {
             backend,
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK

PR #191, opened 2026-07-07
shardtree: Don't panic when checkpoint pruning collapses multiple positions into one leaf

emersonian:
This fixes "Tree state inconsistent with checkpoints." errors on testnet NU6.3 with Ironwood to Ironwood transactions right now.

`LocatedPrunableTree::clear_flags` assumed that when its recursion reaches a `Node::Leaf`, at most one position needs a flag cleared, and panicked with "Tree state inconsistent with checkpoints." otherwise. That assumption does not hold for a note-commitment tree whose frontier advances very little between many checkpoints: a run of checkpointed positions then collapses into a single pruned subtree leaf, and `prune_excess_checkpoints` clears the CHECKPOINT flag for several of those positions in one batch, reaching that leaf with more than one position to clear.

This is reachable in practice for a sparsely-used shielded pool that is still checkpointed at every block by cross-pool checkpoint synchronization (e.g. the Ironwood pool shortly after NU6.3 activation on testnet): its tree accumulates 100+ densely-packed checkpoints on an unmarked single shard, and the first automatic prune past `max_checkpoints` aborts the whole wallet sync.

A collapsed multi-position leaf can only carry CHECKPOINT/REFERENCE retention (a MARKED position is always a level-0 leaf and is never collapsed), so no witness can be lost. Because the leaf carries one retention value shared by every position it covers and `to_clear` cannot tell whether a position sharing the leaf must keep its flag (e.g. a retained anchor), conservatively retain the leaf unchanged rather than panic. The checkpoints are still removed from the checkpoint set by the caller; retaining the leaf flag only defers pruning of an already-pruned leaf, which is safe and self-correcting.

Adds a regression test reproducing the exact panic site.

Relates to #147

### DIFF
--- a/shardtree/src/prunable.rs
+++ b/shardtree/src/prunable.rs
@@ -1156,15 +1156,36 @@ where
                     }
                     Node::Leaf { value: (h, r) } => {
                         trace!("In {:?}, clearing {:?}", root_addr, to_clear);
-                        // When we reach a leaf, we should be down to just a single position
-                        // which should correspond to the last level-0 child of the address's
-                        // subtree range; if it's a checkpoint this will always be the case for
-                        // a partially-pruned branch, and if it's a marked node then it will
-                        // be a level-0 leaf.
+                        // When we reach a leaf, we are usually down to just a single position
+                        // which corresponds to the last level-0 child of the address's subtree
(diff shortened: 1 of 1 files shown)


nuttycom:
Subsumed by #192

PR #192, opened 2026-07-08
shardtree: fix retained-checkpoint pruning corruption and folded-leaf panic

gustavovalverde:
`ShardTree::prune_excess_checkpoints` could corrupt a retained checkpoint's retention flags or panic with "Tree state inconsistent with checkpoints". Two independent defects in how it clears leaf flags.

## Order-dependent flag clearing

Removal candidates skip the retention set, so the checkpoints being removed interleave with retained anchors. The old code cleared flags in one ordered pass. A removed checkpoint iterated after a retained survivor at the same position re-marked the flags the survivor had just preserved, and clearing them folded the survivor's leaf out of the tree.

Repeated checkpoints at one position trigger this fastest, and a stalled source produces exactly that. A tree retaining anchors on a quiet chain hits it within a single pruning window.

The fix uses two passes: mark every flag a removed checkpoint references, then unmark everything a surviving checkpoint still references. Order no longer matters.

## Folded-leaf panic in `clear_flags`

`prune_excess_checkpoints` groups the positions to clear by shard and clears each shard in one call. When a shard has folded a completed subtree into a single annotated leaf, several positions resolve to that one leaf. `clear_flags` assumed a reached leaf always maps to exactly one position, and panicked otherwise.

That assumption is safe. `unite` only folds when the left child is ephemeral and the right carries neither a mark nor a reference, so a folded leaf holds at most the rightmost position's checkpoint flag and never a mark. The extra positions are already-pruned leaves whose flags are gone.

The fix clears the union of the requested flags at the leaf. It is a no-op for the pruned positions and drops the live checkpoint flag as intended.

## Tests

A 400-seed soak covers appends, same-position checkpoint runs, retained and released anchors, and mark removals; it hits the panic without the first fix. A direct test folds positions 0..4 into one level-2 leaf and clears two positions inside it.

### DIFF
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -36,6 +36,13 @@ and this project adheres to Rust's notion of
   - `ShardStore::remove_retained_checkpoint`
   - `ShardStore::retained_checkpoints`
 
+### Fixed
+- `ShardTree::prune_excess_checkpoints` could corrupt a retained checkpoint's
+  leaf retention, or panic with "Tree state inconsistent with checkpoints", when
+  retained anchors interleaved with the checkpoints being pruned. Repeated
+  checkpoints sharing a tree position, as a stalled chain produces, triggered it
+  within a single pruning window.
+
 
 ## [0.6.2] - 2026-02-20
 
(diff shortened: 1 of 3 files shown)


nuttycom (approved):
utACK

PR #196, opened 2026-07-16
shardtree: discard stale state when truncating

nuttycom:
Builds upon #147: this branch includes @gustavovalverde's commits fixing stale cached cap roots on truncation, and extends them to fix two further defects in checkpoint truncation discovered while testing wallet reorg recovery in librustzcash. Merging this PR lands the entire change as a single merge, and GitHub will automatically mark #147 as merged once its commits are reachable from `main`.

The defects fixed on top of #147:

1. **Stale cached annotations survived truncation inside shards.** `LocatedPrunableTree::truncate_to_position` reconstructed the parent nodes along the truncation path with their cached root annotations intact. Such an annotation is a hash over the parent's complete subtree — including the data being truncated away — so a retained annotation caused a spurious `Insert(Conflict(..))` error (or a silently incorrect root) once the truncated region was refilled with different data. Annotations of this shape are routinely written by `insert_frontier`, which wallets invoke with the chain state at the start of every scanned batch, so any wallet that rewinds past a scan-batch boundary and then re-scans a divergent (post-reorg) chain can hit this.

2. **Truncation at a position the tree does not individually represent was silently skipped.** When the truncation position fell in the interior of a pruned subtree (a merged hash node), `truncate_to_position` returned `None`, and `truncate_to_checkpoint` reacted by skipping the truncation while still reporting success. (On shardtree 0.7.0, before #147's fix, the cap had already been truncated at that point, additionally leaving the store internally inconsistent.)

The fix:

- A subtree lying entirely at or below the truncation position is retained unmodified (including its still-valid annotations); parent nodes whose subtrees lose data in the truncation are reconstructed without their cached annotations.
- `truncate_to_position` continues to refuse truncation at a position the tree does not individually represent: succeeding would require discarding retained data along with the requested suffix. `ShardTree::truncate_to_checkpoint` and `truncate_to_checkpoint_depth` now surface that refusal as a `QueryError::CheckpointPruned` error, leaving the store unmodified, instead of reporting success.
- The cap continues to be truncated with #147's `truncate_cache_to_position`: the cap is a cache of hashes recomputable from the shards below, so a cap entry spanning the truncation boundary is discarded rather than making the position unusable. With refusal handled up front, the cap, shard, and checkpoint truncation now proceed unconditionally.

The first commit on top of #147 adds regression tests (red at that commit); the second contains the fix. One of the tests reproduces the wallet-shaped failure entirely through the public API: append leaves, insert a batch-boundary frontier, truncate to an earlier checkpoint, then insert a frontier carrying divergent replacement data — which previously failed with `Insert(Conflict(..))`.

An earlier revision of this change was also validated against librustzcash by path-patching it into its workspace: a wallet-level test that re-mines a transaction on a divergent chain after `rewind_to_chain_state` fails against shardtree 0.7.0 with `PutBlocksCommitmentTree { error: Insert(Conflict(..)) }` whenever earlier scan batches had written frontier ommers above the rewind target, and passes with the fix. (That revision differed from this one only in the handling of non-representable truncation positions; the wallet scenario truncates to a representable position, for which the behavior is identical.)

Supersedes and closes #194.

### DIFF
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -42,6 +42,8 @@ and this project adheres to Rust's notion of
   retained anchors interleaved with the checkpoints being pruned. Repeated
   checkpoints sharing a tree position, as a stalled chain produces, triggered it
   within a single pruning window.
+- Checkpoint truncation now discards cached cap roots that commit to positions
+  beyond the checkpoint, so replacement commitments produce the correct root.
 
 
 ## [0.6.2] - 2026-02-20
--- a/shardtree/CHANGELOG.md
+++ b/shardtree/CHANGELOG.md
@@ -10,6 +10,23 @@ and this project adheres to Rust's notion of
 ### Fixed
 - Checkpoint truncation now discards cached cap roots that commit to positions
   beyond the checkpoint, so replacement commitments produce the correct root.
+- `shardtree::LocatedPrunableTree::truncate_to_position` no longer retains the
+  cached root annotations of the parent nodes it reconstructs along the
+  truncation path. Such an annotation is a hash over the parent's complete
+  subtree — including the data being truncated away — and a stale annotation
+  caused spurious `Insert(Conflict(..))` errors (or silently incorrect roots)
+  when the truncated region was later refilled with different data, e.g. when
+  re-scanning a divergent chain after a reorg.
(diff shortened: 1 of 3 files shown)


nullcopy (approved):
utACK 21a17fa8e4f1e9f5103a529d23cc75204b8f7963

PR #200, opened 2026-07-22
shardtree: tests for retained-anchor survival across pruning

dannywillems:
Adds `shardtree` test coverage for the durable-anchor invariant that
wallet-level anchor retention relies on: a checkpoint marked via
`ShardTree::ensure_retained` is exempt from the `max_checkpoints` prune
budget and survives later checkpoints, with its root and the witnesses for
marked leaves at or before it remaining computable and root-consistent,
including when the retention id is recorded before the checkpoint exists.

### What is added

- **`retained_anchor_survives_pruning`** (property test): over randomized
  layouts (small budget, anchors interleaved with prunable checkpoints,
  shard-boundary crossings), exercising both registration orderings
  (retention recorded before and after the checkpoint exists). Asserts that
  every anchor survives, roots and marked-leaf witnesses stay computable and
  root-consistent, the tip is never pruned, and non-retained survivors never
  exceed the budget.
- **`retained_anchors_survive_single_batch_insert_prune`**: drives the real
  batch path, where `batch_insert` adds every checkpoint and then prunes
  once (the same add-all-then-prune-once shape as `insert_tree`). Anchors
  registered before the batch survive that single prune at, just above, and
  well above the budget. The negative branch shows that without
  pre-registration the deepest anchor is already pruned and retaining it
  afterward cannot bring it back.
- **`retained_anchors_recorded_before_creation_golden`**: golden vectors
  pinning the exact witness path and roots for retention recorded before the
  checkpoint exists, combined with the marked leaf's own checkpoint being
  pruned.

### Context

Follow-up to a review suggestion on librustzcash PR #2728, which fixes
wallet-level anchor retention by registering retention before `insert_tree`
so a scan batch larger than the checkpoint budget cannot prune an anchor
before its retention id is recorded. These tests pin the `shardtree`-level
guarantee that fix depends on.

### Testing

All `shardtree` tests pass; `cargo fmt` (repo toolchain) and
`cargo clippy --all-targets -- -D warnings` are clean.

### DIFF
--- a/shardtree/src/lib.rs
+++ b/shardtree/src/lib.rs
@@ -2327,6 +2327,281 @@ mod tests {
         assert_matches!(tree.store().get_checkpoint(&1), Ok(None));
     }
 
+    #[test]
+    fn retained_anchors_recorded_before_creation_golden() {
+        // Golden vectors for the corner where anchor retention is recorded *before* the
+        // checkpoints exist, and where the marked leaf's own checkpoint is later pruned. Two
+        // anchors (ids 0 and 3) are retained on an empty tree; twelve leaves are then appended,
+        // each its own checkpoint, and the leaf at position 1 is marked. With `max_checkpoints`
+        // of 2, checkpoint 1 — the marked leaf's checkpoint — is pruned, yet the mark, the
+        // retained anchors, and the witness anchored at checkpoint 3 must survive with exact,
+        // known values. `String` hashing here is concatenation, with `_` as the empty leaf, so an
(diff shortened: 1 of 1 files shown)


shardtree/src/lib.rs
```diff
@@ -2327,6 +2327,281 @@ mod tests {
         assert_matches!(tree.store().get_checkpoint(&1), Ok(None));
     }
 
+    #[test]
+    fn retained_anchors_recorded_before_creation_golden() {
```
dannywillems, line 2331: LGTM

